PR #21, opened 2020-04-20
Add PlayerItemCooldownEvent

KennyTV:


BillyGalbreath:
I actually had this event for awhile, but had to remove it because it was causing crashes due to all the upstream changes making everything async (https://github.com/pl3xgaming/Purpur/commit/d4e8659c65ca6b9771277f54b60f5bc35d0a0e61). However, that was when we also had all the Tuinity patches. I'll pull this back in, lets just hope whatever was in Tuinity that caused the crashes hasn't made it way into Paper (or if it has, was fixed).

PR #26, opened 2020-05-07
Move some Goal from Server to Api

HookWoods:
Move the Find Crystal and the Orbit Crystal Goal from Purpur-Server to Purpur-API for the those who use the Mob Goal API and this can be useful to give this goal to mob


BillyGalbreath:
Don't know why this doesn't show as merged, but its merged ^_^ Thanks

https://github.com/pl3xgaming/Purpur/commit/a03667bdd0038c5112ed5121984a28974ee5ab7e

PR #40, opened 2020-08-01
Hide Purpur's configurable seeds in timings reports

chiralpenguin:
This adds Purpur's seed settings to Aikar's set of values to be hidden from timings. Seems to make sense not to expose these as all other generator seeds are hidden, although end spikes aren't a concern there is a possibility knowing Purpur's dungeon seed could be abused from what I can tell.


BillyGalbreath:
Thank you ^_^

Manually merged with changes https://github.com/pl3xgaming/Purpur/commit/d478261f8e15f24edcb4459b72571046faaf0926

PR #45, opened 2020-08-25
Drop raid cooldown patch in favor of bad omen cooldown patch

jpenilla:
The previous patch didn't effectively "break" raid farms, as the player could just loop around the farm until the cooldown wore off, as they would still have bad omen for 1 hour and 40 mins. This patch more effectively breaks the farms when an appropriate cooldown is set (1-5 mins). This patch also adds a config value for the length of the bad omen gotten from killing the patrol leader.


jpenilla:
after writing this patch, I realized this can actually be done quite easily with api.
heres the plugin I made real quick: https://github.com/jmanpenilla/AntiRaidFarm/blob/master/src/main/java/xyz/jpenilla/antiraidfarm/AntiRaidFarm.java

I think the raid cooldown/bad omen cooldown patch can probably be dropped altogether.

jmanpenilla:
After even more testing it seems like the EntityPotionEffectEvent I was trying to use in the plugin is pretty bugged, and this bad omen cooldown patch comes with it's own interesting set of side effects. Because of this I think it's fine to leave in the raid cooldown patch as is (as although it more rate limits the farms than breaks them, it does work consistently.)

PR #51, opened 2020-09-02
Create public_function_for_chatmessage_enderchest.patch

Euphillya:


MrIvanPlays:
use bungee's TranslatableComponent

Euphillya:
> use bungee's TranslatableComponent

what ? But why ? x)
Its just recreate my plugin for all version its stupid, on 1.15/1.16.1 he is public, not private

MrIvanPlays:
so ur plugin doesnt break
mojang did this change so we should respect their decision unless it affects some of the changes bukkit and following projects introduce

Euphillya:
So remove all feactures add by purpur, it's no mojang who has want this feactures : 6 rows enderchest, ridable mobs, 6 rows barrels, modify HP enderdragon

granny:
That's not the correct logic to use in this scenario.

jpenilla:
The feature you want to add is already in the server in the form of the TranslatableComponent API. This API was created for this specific purpose to avoid breakages when Mojang makes internal changes like the ones you mention. Is there a specific reason it doesn't work for you? Example API usage: 
```player.spigot().sendMessage(new TranslatableComponent("container.enderchest"));```

PR #59, opened 2020-09-09
Despawn rate config options per projectile

jpenilla:
Adds a config option in purpur.yml for the despawn rate of each projectile. The default value of -1 will result in unchanged behavior compared to base Paper.


darbyjack:
Ran into this issue today. Would love to see this merged.

jpenilla:
Rebased against ver/1.16.3

I rebuilt patches after editing the purpur config patch. (Added access to SpigotWorldConfig in PurpurWorldConfig, in the same way Paper does it)

PR #70, opened 2020-09-21
Add option to disable zombies pathfinding to villagers only when lagging

chiralpenguin:
I found disabling this feature reduced the impact of zombies on my server by 10-15%, but I do have a lot of zombies and villagers. However, it does break some vanilla gameplay such as the zombification of villagers. This patch uses Purpur's lagging threshold in the same way as the villager brainticks patch to remove the impact on gameplay while the server isn't lagging.

I imagine this could be useful for other survival servers with lots of these entities, it has been really beneficial for my server.

### DIFF
--- a/patches/api/0031-Add-option-to-disable-zombie-aggressiveness-towards-.patch
+++ b/patches/api/0031-Add-option-to-disable-zombie-aggressiveness-towards-.patch
@@ -1,0 +1,19 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: nitricspace <nitricspace@users.noreply.github.com>
+Date: Wed, 23 Sep 2020 22:14:38 +0100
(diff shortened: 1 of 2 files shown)


BillyGalbreath:
The way you've implemented this only checks if the server is lagging when the entity first spawns. I think overriding the a() and b() methods of the Pathfinder you want to disable when lagging and doing the check in there instead would be a better option.

chiralpenguin:
Thanks for the suggestion Billy, that's a much better implementation. I've overridden the methods in an anonymous inner class as you suggested on Discord. Since it's an anonymous class, I've also added the new mob goals to MobGoalHelper's deobfuscation map.

BillyGalbreath:
Manual merged by accident :3

https://github.com/pl3xgaming/Purpur/commit/17d8ee835bd53b8e38a475b8bb62e740572f6dee

Issue #66, opened 2020-09-18
[Feature Request] AFK Sugguestions


fredagsguf:
Hi, i think the AFK system is lacking a bit - don't get me wrong its really nice and also nice to hook into, but i'd like more options e.g an option to make it change the tab suffix of a player instead of tab prefix.
maybe make it sort all AFK lowest on scoreboard, could be done by hooking into nametagedit?

Maybe also a option to execute commands when a player is set as afk?

Not sure i'm asking the right place but oh well...

BillyGalbreath:
The AFK system isn't meant to be fancy. I've included the needed plugin API to the system for anyone to enhance as they see fit by using a plugin

PR #72, opened 2020-09-24
Add more configuration options to the AFK API.

chiralpenguin:
Discussed exactly what AFK configuration options were being requested with @Laukage, who opened https://github.com/pl3xgaming/Purpur/issues/66,  on [Discord](https://discordapp.com/channels/685683385313919172/720374890472931328/758682171832664186) and came up with this set of what I thought would be reasonable additions.

Just created this PR so you could pick and choose and additions to the AFK system that you think should be included in Purpur, if any.


BillyGalbreath:
See comment on #66

PR #80, opened 2020-10-21
port async advancements patch to purpur

Titaniumtown:
Just a note: I'm such a noob at this, so if this looks super bad or something, tell me.

I ported the async advancements patch from https://github.com/PaperMC/Paper/pull/4160 to Purpur. 

But it doesn't seem to build and I can't figure it out. The error always is: 
```
Caused by: org.eclipse.aether.transfer.ArtifactNotFoundException: Failure to find co.aikar:cleaner:jar:1.0-SNAPSHOT in https://hub.spigotmc.org/nexus/content/groups/public/ was cached in the local repository, resolution will not be reattempted until the update interval of spigotmc-public has elapsed or updates are forced
```
I can't seem to figure out how to fix that, so any help would be appreciated!


Titaniumtown:
I've now fixed the issue I was having. It builds for me. I'll convert this from a draft PR.

Titaniumtown:
I did some testing w/o any issues. Advancements work as expected.

Huge-mistake:
Wow, is this a merger?

granny:
> Wow, is this a merger?

No. The referenced PR in that commit is from Pufferfish.

PR #87, opened 2020-10-31
Fix client lag on advancement loading

jpenilla:
When new advancements are added via the UnsafeValues#loadAdvancement
API, it triggers a full datapack reload when this is not necessary. The
advancement is already loaded directly into the advancement registry,
and the point of saving the advancement to the Bukkit datapack seems to
be for persistence. By removing the call to reload datapacks when an
advancement is loaded, the client no longer completely freezes up when
adding a new advancement.

edit: I also opened this PR upstream, so don't merge for now


BillyGalbreath:
Needs to be behind a config option that defaults to vanilla behavior (just in case). Then I'll merge ^_^

Issue #85, opened 2020-10-26
Feature request: Adjustable breeding cooldown for breedable mobs


Titaniumtown:
Could a config option be added that allows the modification (in the form of a config option ofc) of the cool down you have to wait until you can breed the animals again? This should most likely be a per-mob config option.

PR #92, opened 2020-11-13
Add adjustable breeding cooldown to config

montlikadani:
This patch literally not the best implementation, because the cooldown is implemented with NBTs, but if we want this, we should extend the EntityAnimal to PersistentBase. I'm unsure we can do this way. Also this option is not per-mob handled, it is for global, but for each player.

Related #85


Titaniumtown:
Thanks for addressing my patch request!

BillyGalbreath:
I like it

Made some small adjustments (hope you dont mind)

I'm wondering if we should make the Map static to cover per animal _type_ rather than per animal. Thoughts?

montlikadani:
Yea, I also think _static_ would be better. You can merge if you pushed the changes. ✌️

montlikadani:
Closing, as there is a better implementation ⬆️

PR #94, opened 2020-11-14
Option to use farmland mechanics from Alpha/Beta minecraft

Yive:
Updated it to also work with stone walls due to it not being a thing back then.


jpenilla:
Does `BlockCobbleWall` cover all the different types of walls?

Yive:
Yes it does cover every wall and every future walls they plan to add, same with fences.

PR #95, opened 2020-11-15
Breeding configuration patches

jpenilla:
This PR includes two patches (willing to split it up if requested)
- My updated version of the patch from #92
  - Made the cooldowns per animal type
  - Store the cooldowns per world instead of per entity (because the config option is per world)
  - If the player has a cooldown for an animal type, don't let them put that animal into love mode/don't consume the breeding item
- Make-animal-breeding-times-configurable.patch
  - Makes the waiting time for breeding animals configurable per animal type, instead of always the vanilla 5 minutes.
  - Note: currently the config includes some animals which are technically not breedable, because I simply added a config for all entity types that are `Animals`. This is easily fixable by making the code for the config a bit more verbose.
    img

Also related: #85

### DIFF
--- a/patches/server/0140-Make-animal-breeding-times-configurable.patch
+++ b/patches/server/0140-Make-animal-breeding-times-configurable.patch
@@ -1,0 +1,639 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: jmp <jasonpenilla2@me.com>
+Date: Sun, 15 Nov 2020 02:18:15 -0800
+Subject: [PATCH] Make animal breeding times configurable
+
+
+diff --git a/src/main/java/net/minecraft/server/EntityAnimal.java b/src/main/java/net/minecraft/server/EntityAnimal.java
+index 0b3d0c84f..e0402475c 100644
+--- a/src/main/java/net/minecraft/server/EntityAnimal.java
++++ b/src/main/java/net/minecraft/server/EntityAnimal.java
+@@ -13,6 +13,7 @@ public abstract class EntityAnimal extends EntityAgeable {
+     public int loveTicks;
+     public UUID breedCause;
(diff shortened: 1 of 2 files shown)


patches/server/0140-Make-animal-breeding-times-configurable.patch
```diff
@@ -0,0 +1,64 @@
...
++import org.bukkit.entity.Animals;
++import org.bukkit.entity.EntityType;
+ 
+ import java.util.ArrayList;
+ import java.util.HashMap;
+@@ -88,6 +91,19 @@ public class PurpurWorldConfig {
+         armorstandStepHeight = (float) getDouble("gameplay-mechanics.armorstand.step-height", armorstandStepHeight);
+     }
+ 
++    public Map<EntityTypes<?>, Integer> animalBreedTimes = new HashMap<>();
```
montlikadani: Please add `final`

katekerllenevich: dumb question, but when you reload, wouldn't it have to not be `final`? or am I over thinking it lol

montlikadani: No, this field shouldn't be modified by any of code, because it will breaks the added EntityTypes. Every of other options will be loaded in PurpurConfig and PurpurWorldConfig.

katekerllenevich: ahhh thanks lol

BillyGalbreath:
I feel like these should be split into their appropriate `mobs.entity_type.` options. It makes for more work on the coder, yes, but it keeps things organized for the end user. Thoughts?

jpenilla:
> I feel like these should be split into their appropriate `mobs.entity_type.` options. It makes for more work on the coder, yes, but it keeps things organized for the end user. Thoughts?

Agreed. I'll update the PR

PR #101, opened 2020-11-30
Add configurable mob left handedness chance.

katekerllenevich:
Just a fun quirky patch, that I thought was pretty funny :P

Let me know if I need to change the config location.


Titaniumtown (approved):
Fun idea!

PR #104, opened 2020-12-02
[ci-skip] Update readme with downloads API info

jpenilla:
Added some info about the downloads API to the readme, and reverted the readme to use the Purpur logo instead of the website banner. In my opinion the logo looks much cleaner on the white background of GitHub.


katekerllenevich (approved):
looks good to me!

I agree that the square logo looks better in the readme.

PR #116, opened 2020-12-10
Configurable default wolf collar color

encode42:
Configurable via `world-settings.<world>.mobs.wolf.default-collar-color`.
Takes EnumColor as an input. Invalid inputs default to RED.


katekerllenevich:
how about naming it `default-collar-color` so it doesn't seem like that will be the only tamed collar color

jpenilla:
> how about naming it `default-collar-color` so it doesn't seem like that will be the only tamed collar color

Yeah I agree, and also name the patch accordingly "Configurable default wolf collar color"

katekerllenevich (approved):
lgtm!

PR #139, opened 2021-01-09
Add unsafe Entity serialization API

Proximyst:
This is in the `UnsafeValues` interface which has no guarantee of stuff staying around or even working. This is an intentional, concious choice, and it's up to the merger to determine if they want it to be on `Entity`, where the lack of that very concious guarantee is gone.


patches/api/0040-Add-unsafe-Entity-serialization-API.patch
```diff
@@ -0,0 +1,61 @@
...
+index 97b9ade0e771eae663fb42f91e15545034d58fc9..a4db2cc96a52e785bf7cfce6d16a770f8d0ae7e2 100644
+--- a/src/main/java/org/bukkit/UnsafeValues.java
++++ b/src/main/java/org/bukkit/UnsafeValues.java
+@@ -135,4 +135,9 @@ public interface UnsafeValues {
+     public int nextEntityId();
+ 
+     // Paper end
++
++    // Purpur start
++    byte[] serializeEntity(org.bukkit.entity.Entity entity);
```
jpenilla: Even though these are "unsafe" APIs, I'd prefer for them to have Javadoc

PR #140, opened 2021-01-09
Movement options for armour stands

Proximyst:
This is not a "armour stand lag no more" patch, however it may heavily reduce the impact of them if used in a malicious way. This is useful for anyone targeted by ~~nolife~~ malicious groups intending to cause a DoS through lagging the server to a halt.

### DIFF
--- a/patches/server/0171-Movement-options-for-armour-stands.patch
+++ b/patches/server/0171-Movement-options-for-armour-stands.patch
@@ -16,32 +16,21 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
-diff --git a/src/main/java/net/minecraft/server/Block.java b/src/main/java/net/minecraft/server/Block.java
-index c276a2896ba73d86e7e18d656903b36d2970ac41..15ec5b8d41509ca886cb699cdcb5b45361b043d6 100644
---- a/src/main/java/net/minecraft/server/Block.java
-+++ b/src/main/java/net/minecraft/server/Block.java
-@@ -44,6 +44,19 @@ public class Block extends BlockBase implements IMaterial {
-         return timing;
+diff --git a/src/main/java/net/minecraft/server/Entity.java b/src/main/java/net/minecraft/server/Entity.java
+index 073be230a8c0e2ccf6b4304e2988c270749f1a35..b521a9984c327f74bccd634a8bfe26f2cd638462 100644
+--- a/src/main/java/net/minecraft/server/Entity.java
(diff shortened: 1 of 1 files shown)


patches/server/0171-Movement-options-for-armour-stands.patch
```diff
@@ -0,0 +1,99 @@
...
+         this.armorItems = NonNullList.a(4, ItemStack.b);
+         this.headPose = EntityArmorStand.bj;
+@@ -897,4 +899,18 @@ public class EntityArmorStand extends EntityLiving {
+         return true;
+     }
+     // Paper end
++
++    // Purpur start
++    @Override
++    void aL() {
```
katekerllenevich, line 56: What does aL() do? It probably needs and OBFHELPER.

Proximyst: How do you propose that without breaking potential stupid plugins that call the method? It wouldn't be possible to just rename. Best I'd come up with is just to put a comment on the super with the Mojang name, so it will conflict on change.

patches/server/0171-Movement-options-for-armour-stands.patch
```diff
@@ -0,0 +1,99 @@
...
+index c276a2896ba73d86e7e18d656903b36d2970ac41..15ec5b8d41509ca886cb699cdcb5b45361b043d6 100644
+--- a/src/main/java/net/minecraft/server/Block.java
++++ b/src/main/java/net/minecraft/server/Block.java
+@@ -44,6 +44,19 @@ public class Block extends BlockBase implements IMaterial {
+         return timing;
+     }
+     // Paper end
++    // Purpur start
++    public final boolean isFence() {
++        return this == Blocks.ACACIA_FENCE ||
```
jpenilla: why not `instaceof BlockFence`? this seems more likely to be missed if new fences are added

Proximyst: Because I didn't know of it 😝 Will do that shortly.

patches/server/0171-Movement-options-for-armour-stands.patch
jpenilla: I would prefer to have the comment on the super so it does conflict on change. Could do some weird stuff like
```java
    // Purpur start - OBFHELPER
    void aL() { this.namedMethod(); }
    void namedMethod() {
    // Purpur end
```
but adding a comment to the super seems like a more sane way to ensure we notice a change

Proximyst: I've done a comment for now (didn't see your comment before doing it, oops).

PR #185, opened 2021-03-05
One Punch Man!

katekerllenevich:
Ported this mod by @Fourmisain: https://github.com/Fourmisain/CreativeOnePunch

If you are in creative and you punch someone while not holding anything. It dies.


Fourmisain:
Oh, surprised to see this here!

If I can criticize this implementation a bit:
This will only one punch entities when you're not holding anything, which makes sense given the name.
However, practically speaking, when working in creative your hotbar will always be full which makes this feature inconvenient to use.
This is why I originally checked if the player is not holding a weapon (by which I mean any item with an attack value), so you can one punch even with building blocks or similar in your hand - and you can still test the damage of your weapons on mobs.

katekerllenevich:
Oh... I must have read it wrong. I'll definitely PR a fix.

PR #201, opened 2021-03-17
Config to make Creepers explode on death

### LINKED ISSUE #200: (Suggestion) Creepers explode when they die
Used to be in the alpha versions of Minecraft; whenever killing a creeper they would explode, so creepers still would have the same behavior just whenever they die, they still explode. A nice addition to difficulty especially combined with charged creepers.

encode42:
Brings back an alpha mechanic of making Creepers explode when killed by players.
Example

Resolves #200


encode42:
Temporarily a draft as I improve the order everything happens.

PR #188, opened 2021-03-06
Add bookshelf containers (close #169)

katekerllenevich:
This is a WIP PR for adding containers to bookshelves as requested by #169.

Currently this only gives bookshelves their own class. Earlier testing gives this weird effect where it instantly closes the container upon opening (Using [this code](https://paste.gg/p/anonymous/1d0af220c35d4d95a0380e7e2392ebd1)) the console would also spam `[08:34:54 ERROR]: Block at -24, 66, 34 is Block{minecraft:bookshelf} but has net.minecraft.server.TileEntityDispenser@4ab2e39e. Bukkit will attempt to fix this, but there may be additional damage that we cannot recover.`.

PR #207, opened 2021-03-17
Add bookshelf containers (close #169)

### LINKED ISSUE #169: (Suggestion) Make Bookshelves into usable containers for books and alike
Something similar to this https://www.spigotmc.org/resources/bookshelf-enchantment-table-boosting-supports-protection-plugins-pistons-hoppers-droppers.75804/ would be pretty hecken cool.

katekerllenevich:
Continuation of #188.
New issues: The server never thinks that it should have bookshelf containers cause the values get init-ed before the config does cause its static.


katekerllenevich:
Ok, I worked out a solution for the issue and everything works now. Except for one issue, whenever you start the server (option enabled or not) you get this message: `No data fixer registered for bookshelf`. Other than that, this PR is ready.

katekerllenevich:

BillyGalbreath:
After discussing this it has been decided that this is just too big of a change to the server code that leads to numerous issues. It is best to just handle this feature in a plugin which will be much more optimized, much more configurable, and with far less code.

Issue #206, opened 2021-03-17
[Suggestion] Villager trade disabler


Misio12320:
Add support to disable trades for villager even when they have a profession. I suppose it would be cool to stop trading on a server via purpur.yml config. 😳

BillyGalbreath:
https://github.com/pl3xgaming/Purpur/commit/5f76cb35ce51ac5e2230ac7f116bfec66d7ceba0

PR #210, opened 2021-03-18
Add a config for villager trading (add #206)

katekerllenevich:
Allows server owners to disable trading completely (#206)


katekerllenevich:

PR #215, opened 2021-03-18
Add config for infinity on crossbows

katekerllenevich:
Allows infinity to naturally be placed on crossbows from the enchantment table and allows the crossbow to use infinity.

Only issue is that when loading the crossbow the arrow disappears because of client desync.


katekerllenevich:
* but the arrow re-appears after firing

katekerllenevich:

PR #230, opened 2021-03-24
Rename no-tick patch/options to no-random-tick

encode42:
This will hopefully reduce confusion regarding this option.


encode42:
Currently needs a way to add to a list in the configuration file.

encode42:
Should be ready for merge!

PR #278, opened 2021-04-11
Fix MyPets compatibility Ride Mobs

Euphillya:


katekerllenevich:
this is redundant, as `entity.hasRider()` checks `rider != null`

Euphillya:
Theoretically this is the case but with the check of the rider != null, console spam does not exist anymore

encode42:
This isn't a proper fix, here's what Billy had to say about the fix being applied to Yatopia a few months ago:

> [mentions ishland] your "fix" for mypet isn't the proper fix. It's just masking the problem a little further (if it even fixes the problem at all, I haven't tested it). The plugin is ticking the entity async, which is why the rider become null between 2 instructuon sets (were talking sing digit nanoseconds here). The proper fix is for the plugin to stop ticking entities async...
> ```java
> if (entity.hasRider()) {
>     tick(entity.getRider());
> }
> ```
> For `entity.getRider()` to be null, there _has_ to be another thread running and removing the rider at that precise moment. `entity.hasRider()` can't be not null and then in the next instruction set be null.. something happened in the nano seconds between there. i could have easily "fixed" it the way you have, or even added a null check to tick(), but thats just masking the problem and preventing tick() from ever running anyways because something else is screwing with the rider

This is the actual fix that must be done on MyPet's side: https://github.com/xXKeyleXx/MyPet/issues/1425#issuecomment-775284641

Euphillya:
Ah ok after i am just look error in my console and try patch this ^^

PR #292, opened 2021-04-23
Migrate multithreaded tracker config

encode42:
This isn't an Airplane feature anymore and causes confusion for the Airplane support team and Purpur's users.

- AIR doesn't have a method to set a value as null. We can either:
    * Set section comments to reflect migration [current solution]
    * Utilize an overcomplicated method to remove the section
    * Do nothing [Airplane's solution]
- Setting the section comment seems to be broken currently. I contacted Paul for support on why it's not working. I'd like to get this fixed before merge.
- This is a new patch since I don't want to deal with all the merge conflicts placing it after the Purpur config files. Feel free to adjust this if anyone has the time.

### DIFF
--- a/patches/server/0005-Purpur-config-files.patch
+++ b/patches/server/0005-Purpur-config-files.patch
@@ -86,7 +86,7 @@ index 51e6cd6119465f9fd6385072997971449afb5f42..f8261e21a84bf8c29d72116fc3166dc7
          this.ticksPerAnimalSpawns = this.getServer().getTicksPerAnimalSpawns(); // CraftBukkit
 diff --git a/src/main/java/net/pl3x/purpur/PurpurConfig.java b/src/main/java/net/pl3x/purpur/PurpurConfig.java
 new file mode 100644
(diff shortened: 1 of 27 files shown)


YouHaveTrouble (approved):
lgtm

Titaniumtown (approved):

12emin34 (approved):
👍

BillyGalbreath:
Now you see why I haven't gotten around to doing this myself. To do a proper migration airplane needs a method so I can save a null value to (or remove altogether) these 2 options. Until that exists something like this isn't going to happen. And when it does happen it will be a complete clusterfuck because I'm going to move the entire patch to the end of the list.

On the other hand, I'm not all that pressed because its a very minor QoL change that has no real benefit, especially seeing that the option is disabled by default (i.e., most people wont even see the option exists). Mojang is going to release 1.17 here in the next few months anyways, and this is one of the patches that are being dropped at that time.

I only kept this patch around because airplane decided to drop it entirely, even though it works fine for most servers. It is not my intention to maintain it forever.

PR #294, opened 2021-04-23
Add Velocity in Online Mode.

Euphillya:
Bungee was added, why not velocity too


BillyGalbreath:
`PaperConfig.isProxyOnlineMode()` already checks for velocity, too. The bug i think you are trying to fix is in the Timings Viewer code, not the server software.

Euphillya:
Yes its my idea of this patch, i am not verry check sorry

Roan-V:
https://github.com/aikar/timings/commit/63fcbeb2772022da8dca5def7ac400cc1e1d49f1
How is this related then? It seems like support has already been added

BillyGalbreath:
> [aikar/timings@63fcbeb](https://github.com/aikar/timings/commit/63fcbeb2772022da8dca5def7ac400cc1e1d49f1)
> How is this related then? It seems like support has already been added

My fork of the Timings Viewer is way older than that commit.

Roan-V:
So you should update :P

BillyGalbreath:
That's the plan. Eventually ;)

PR #302, opened 2021-04-29
Added Information in Hives when broken

Euphillya:
Allows you to add the number of bees to the honey level of the hive.

image


encode42:
This would be better off as a plugin. If merged it'll lead to a never ending chain of requests to add more and more NBT data to item lore. Additionally, accomplishing this with a plugin would be easier to maintain and customize with less lines of code. This is helpful information for the player to know though 👍

PR #304, opened 2021-04-30
Config for health to impact Creeper explosion radius

encode42:
This allows for a Creeper's explosion to be affected by its health value.
This should scale with any custom explosion radius or max health set.
Equation: Creeper's explosion radius multiplied by percentage of max health `(health / max health) * (explosion radius * charged multiplier)`

Update for those wondering why this hasn't been merged yet: We're currently making sure the JavaScript engine Purpur integrates is functioning correctly, then this PR will change to allow for the equation to be customized by the user.

**Explosion radius with feature disabled:**

**Explosion radius at full health:**

**Explosion radius at half health:**


katekerllenevich (approved):
lgtm, a cool feature would be to invert this and make the explosions bigger with less health.

YouHaveTrouble:
> lgtm, a cool feature would be to invert this and make the explosions bigger with less health.

Yeah, I suggested making config option a number value that multiplies explosion force and if it's negative it reverses the health scaling calculation

katekerllenevich:
we could use the js expression evaluation to make the full equation customizable

PR #313, opened 2021-05-03
Updated Upstream (Paper)

giovanniwijaya:
Upstream has released updates that appear to apply and compile correctly

Paper Changes:
976c6d425 [CI-SKIP] [Auto] Rebuild Patches
51deec726 Drop unneeded portal patch (#5566)
07a18c457 [Auto] Updated Upstream (CraftBukkit)
6733d7875 [Auto] Updated Upstream (Bukkit)


BillyGalbreath:
Thanks, but we don't need a PR every time something happens upstream. We got it covered ^_^

granny:
Keep in mind you don't need to refork Purpur everytime you want to send in a PR. The best way to PR something in the future is by creating a new branch and adding your changes there.

PR #314, opened 2021-05-03
Configurable enchantment max level

Euphillya:
This patch is from an old feature of Purpur:
https://github.com/pl3xgaming/Purpur/blob/ver/1.15/patches/server/0103-Configurable-enchantment-max-level.patch


BillyGalbreath:
This patch was dropped because it did more harm that good. Too many people tried setting values that broke the balance of the game in unexpected ways. It wasn't worth the trouble trying to support that.

PR #315, opened 2021-05-03
Allows you to force when you want to be halloween and config the luck like the drop.

Euphillya:


Euphillya:
I think I can do better than that, I would do a PR again at this precise moment.

PR #321, opened 2021-05-07
Extends Halloween Airplane Patch

Euphillya:
Airplane allowed to check the beats if it was halloween season, I extended the concept to halloween affected entities.

### DIFF
--- a/patches/server/0207-Extends-Check-halloween-Airplane-Patch.patch
+++ b/patches/server/0207-Extends-Check-halloween-Airplane-Patch.patch
@@ -1,0 +1,114 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: DoctaEnkoda <bierquejason@gmail.com>
+Date: Fri, 7 May 2021 03:35:42 +0200
+Subject: [PATCH] Extends Check halloween Airplane Patch
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/ambient/EntityBat.java b/src/main/java/net/minecraft/world/entity/ambient/EntityBat.java
+index 88a0d51f78bee9a1b0762b46db4d91a9f8f8c75d..2af1fa1013899176469dbd78774ce5d51eb4234f 100644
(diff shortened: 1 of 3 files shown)


BillyGalbreath (approved):

patches/server/0207-Extends-Check-halloween-Airplane-Patch.patch
```diff
@@ -0,0 +1,83 @@
...
+index 88a0d51f78bee9a1b0762b46db4d91a9f8f8c75d..5284f8831baf6f5a2b04eb304eb8357851d279c7 100644
+--- a/src/main/java/net/minecraft/world/entity/ambient/EntityBat.java
++++ b/src/main/java/net/minecraft/world/entity/ambient/EntityBat.java
+@@ -311,17 +311,31 @@ public class EntityBat extends EntityAmbient {
+     private static boolean isSpookySeason = false;
+     private static final int ONE_HOUR = 20 * 60 * 60;
+     private static int lastSpookyCheck = -ONE_HOUR;
+-    private static boolean eJ() {
++    // Purpur start
++    private static boolean isHalloween = false;
```
PaulBGD: Main issue with this modification imo is there's no reason for the spooky season stuff to be inside the bat class..

PaulBGD (changes requested):

patches/server/0207-Extends-Check-halloween-Airplane-Patch.patch
BillyGalbreath: I agree, but we didn't put it there ;) This is just an extension of existing code

Euphillya: In fact when I started to make the patch, I had the idea to use the method in the BAT class but as for the bat it's a period and not the day of Halloween, I I had the idea of ​​extracting it in another file and joining the 2 methods if one day mojang, or an extension of Purpur or other wants to add functionalities to its events of the year, it is possible.

PR #322, opened 2021-05-07
Update the "changing the blocks that turn into paths" defaults to reflect the vanilla defaults.

Roan-V:


Euphillya:
Hello, I don't understand why you removed the dirt block?

Roan-V:
Because minecraft doesn't do that in vanilla https://minecraft.fandom.com/wiki/Grass_Path

This is just an edit of the default config options, you can obviously still add dirt to that list on your own server ^_^

12emin34:
The dirt block was likely accidentally left in when Billy made the list configurable

encode42 (changes requested):
Coarse dirt and mycelium can only be used to make paths in 1.17, 1.16 only has grass

Roan-V:
Ah you're right, I just tested it
The wiki lied to me >:(

encode42 (approved):

encode42:
Keep in mind that you can't edit patches by hand or else the index doesn't match up, patches must be edited from the source

PR #333, opened 2021-05-14
Configurable powered rail boost modifier

### LINKED ISSUE #329: (Enhancement suggestion) - Powered Rails configurable speed boost
For ages I've dreamt of Mojang making Minecarts & rails have any sort of relevance in modern Minecraft other than uses in redstone contraptions. Long distance travelling is just not worth the effort of creating the track for anymore when it travels at a fraction of the speed of ice boat tracks or Elytras.

However I hear that this is something that can theoretically be implemented by a server jar that cares about adding relevant and cool options, and so I've ended up here after being advised to add my suggestion.

Things that I can imagine could cause issues is that this option would likely ruin Minecart with Hoppers for redstone contraptions and other redstone-related uses of minecarts.
The only thing I can imagine would fix this would be if you could someone control how it'd reach a new configured speed. Such as requiring the minecart to have travelled over a certain amount of powered rails before it starts applying the new speed boost, to prevent the new speed limit of affecting any redstone contraptions as I imagine few use longer tracks of only powered rails.
So for example you could need it to have travelled 16 blocks on powered rails before it starts ramping up the speed further.

This suggestion on how it could be fixed comes from someone who has zero experience developing stuff for real, and I really have no clue of the limitations there are so do take all I've said with a grain of salt.

BomBardyGamer:
Closes #329 

This adds an option to change the speed boost that minecarts gain from hitting a powered rail.

(Note: This doesn't work with furnace minecarts because they're weird :))

### DIFF
--- a/patches/server/0215-Configurable-powered-rail-boost-modifier.patch
+++ b/patches/server/0215-Configurable-powered-rail-boost-modifier.patch
@@ -18,7 +18,7 @@ index 7b49544210d087f5006a83c2a0d5c47c785c567f..9e91fc85e8f7ebadde239941700282fd
                  Vec3D vec3d5 = this.getMot();
                  double d21 = vec3d5.x;
 diff --git a/src/main/java/net/pl3x/purpur/PurpurWorldConfig.java b/src/main/java/net/pl3x/purpur/PurpurWorldConfig.java
-index db8b31972b38f907a53b4f63465c83d1f47f4dc3..86af0800d86c486bf05d151d80286e1e2c34e90c 100644
+index db8b31972b38f907a53b4f63465c83d1f47f4dc3..99c196f1262660b901f84c0ac66d10c9d1e6bcf7 100644
 --- a/src/main/java/net/pl3x/purpur/PurpurWorldConfig.java
 +++ b/src/main/java/net/pl3x/purpur/PurpurWorldConfig.java
 @@ -133,6 +133,7 @@ public class PurpurWorldConfig {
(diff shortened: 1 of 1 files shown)


patches/server/0215-Configurable-powered-rail-boost-modifier.patch
```diff
@@ -0,0 +1,39 @@
...
+     public Map<Block, Double> minecartControllableBlockSpeeds = new HashMap<>();
++    public double poweredRailBoostModifier = 0.06;
+     private void minecartSettings() {
+         if (PurpurConfig.version < 12) {
+             boolean oldBool = getBoolean("gameplay-mechanics.controllable-minecarts.place-anywhere", minecartPlaceAnywhere);
+@@ -185,6 +186,7 @@ public class PurpurWorldConfig {
+             set("gameplay-mechanics.minecart.controllable.block-speed.grass_block", 0.3D);
+             set("gameplay-mechanics.minecart.controllable.block-speed.stone", 0.5D);
+         }
++        poweredRailBoostModifier = getDouble("gameplay-mechanics.minecart.powered-rail-boost-modifier", poweredRailBoostModifier);
```
granny:
```suggestion
+        poweredRailBoostModifier = getDouble("gameplay-mechanics.minecart.powered-rail.boost-modifier", poweredRailBoostModifier);
```

granny (changes requested):
LGTM, just one minor change

granny (approved):

katekerllenevich (approved):

PR #334, opened 2021-05-15
Remove Warning Java Version When Server Start

Euphillya:


katekerllenevich:
I'm against this change. This message only shows on soon to be unsupported Java, there should be no "off" for this other than updating.

Euphillya:
I thought of turning it off because currently the concern is that there are plugins that still do not work on Java 16 because they use some currently removed methods.

Draycia:
The warning has no effect on plugins and is merely that - a warning. 
If any plugins don't work on j16 the issue should be reported and resolved, instead of ignoring the issue and pretending it doesn't exist.

granny:
Exactly what @Draycia said. Closing for that reason

PR #348, opened 2021-05-19
Add config for the unverified username message

katekerllenevich:
I need someone to test this as I don't have any cracked accounts.


encode42:
Works as expected for me

PR #361, opened 2021-05-24
Add Config Shulker Box 6 rows

Euphillya:
https://github.com/pl3xgaming/Purpur/issues/24


BillyGalbreath:
Have you actually tested this thoroughly? This is something I've tried to do before, but it just has too many problems. You can place shulker boxes inside rows 4-6, but only client side. The data just get too large in some circumstances and can crash clients. Plugins that toy with shulker boxes or their inventories get hung up and create even more problems. Its a huge mess. There's other stuff, too, I just dont remember them all. I just remember there were too many problems with how Mojang implemented these containers with the client. I'm going to have to say no to this, yet again, because its just too unsafe.

PR #377, opened 2021-06-04
Wolves can attack more entity

Euphillya:


granny:
I'm not really sure what you're going for here. What does this patch do? What are you trying to accomplish? Would love to see an explanation of what you're going for in this and future PRs.

Euphillya:
Currently, wolves only attack when they see a skeleton, it seems logical in itself since it is a bag of bones. But then I think once tamed the wolf can be a good support during the PVE and doesn't wait for the player to get attacked for the wolf to attack. Making it aggressive on certain entities or even players will improve the gameplay for combat but at the same time make it more difficult if the wolf becomes hostile.

granny:
Then I feel the option should be renamed to maybe `mobs.wolf.tamed-auto-aggression` or something similar. It'd also be a better idea to have it be a list of mobs instead of hardcoding them to multiple options.

encode42 (changes requested):
You can go one step further by making the entities to aggro customizable via a list (as said above)

Euphillya:
I will re-work the Pull Request this week, I was not at home.

Euphillya:
I close the request while waiting for 1.17 to add the new mobs and possibility.

PR #389, opened 2021-06-12
Config for allow water in nether

Euphillya:
Extends patch 0082


granny:
won't be looked at until after a stable release of 1.17

Issue #393, opened 2021-06-16
Permission to let certained players join when slots are full


emielderckx:
## Is your feature request related to a problem?
I want to let staff and content creators still be able to join my server when its full (slots)

## Describe the solution you'd like.
I would love a permission that let people still join the server.

## Describe alternatives you've considered.
I know some plugins have this feature, but I really don't want to add essentials only for this and it sounds like a really good Purpur feature. 

Also Roan on Discord said that this is possible for OP's but I don't really want to give content creators and all staff op so thats not an option.

## Agreements
Please agree to the following.
- [x] I have searched for and ensured there isn't already an open issue regarding this.
- [x] I have ensured the feature I'm requesting isn't already in the latest supported Purpur build.

Roan-V:
In extension of this
It would be cool if basically anything that requires op could be turned into a permission
E.g. using structure blocks

PR #394, opened 2021-06-16
Allow player join full server by permission

Euphillya:
#393


granny:
won't be looked at until after a stable release of 1.17

PR #395, opened 2021-06-18
Updated Upstream (Paper)

Euphillya:
Upstream has released updates that appear to apply and compile correctly

Paper Changes :
2db85c5 fix: origin world can be unknown while knowing location
903fa9a Enable chunk load for feature search by default

Purpur Change :
Change submodule branch master to ver/1.16.5 Paper


katekerllenevich (approved):
might as well for our (hopefully) final build of 1.16.5

PR #408, opened 2021-06-24
Add Tuinity 1.17 upstream

PaulBGD:
Leaves in independent updating of either Paper/Tuinity.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -71,4 +71,19 @@ paperweight {
             serverOutputDir.set(layout.projectDirectory.dir("Purpur-Server"))
         }
     }
+
+    useStandardUpstream("tuinity") {
+        url.set(github("Tuinity", "Tuinity"))
+        ref.set(providers.gradleProperty("tuinityCommit"))
+
+        withStandardPatcher {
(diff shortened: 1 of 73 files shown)


PaulBGD:
Hmm while it builds fine locally before I cleared caches, it seems Paperweight isn't quite comfortable with multiple upstreams yet. I'm looking into the issue a bit in Paperweight, but it might be better for now to go back to squashing Tuinity to a commit.

BillyGalbreath:
I believe its because of the original issue I mentioned a long time ago about how Tuinity and Paper often conflict with each other because of the long delays between Tuinity's upstream merges.

I am testing something locally where I maintain an up-to-date fork of Tuinity and seeing if paperweight likes that any better. If it pans out I'll run with it.

PaulBGD:
DenWav says this is a supported feature, I presume the way it would work is the project which uses multiple upstreams solves conflicts as its own patch.

I assume maintaining a separate updated fork of Tuinity would work just fine (since it's linear again) but this doesn't seem like a huge bug so it should be fixable once I can figure out where it is.

PR #412, opened 2021-06-25
Option to make drowned break doors

12emin34:
I noticed that drowned don't break doors, so why not add an option to enable it


katekerllenevich (approved):
just needs to be rebased

PR #417, opened 2021-06-26
Shulker spawn from bullet options

BillyGalbreath:
This allows changing options regarding the new 1.17 feature of spawning a new shulker when one gets hit by a shulker bullet.

I am opening this as a PR to get community feedback.

Vanilla behavior:

When a shulker gets hit by a shulker bullet it will teleport to a new position and spawn a new shulker of the same color where it originally was if these conditions are met:

* The shulker has over 50% of its health, or it passes a RNG check (75%)
* The shulker lid is open
* The shulker finds a place to teleport to
* It passes a RNG check based on the number of shulkers around it in a 17x17x17 area (100% chance, minus 20% for each additional spawner within range)

The new options allow changing:

* Whether or not the shulker lid needs to be open
* The base RNG chance (default 100%)
* The range of the nearby shulkers
* The equation used to decrease the base chance according to the number of nearby shulkers
* Whether the new shulker will be a random color or the same as the teleported one was

### DIFF
--- a/patches/server/0215-Shulker-spawn-from-bullet-options.patch
+++ b/patches/server/0215-Shulker-spawn-from-bullet-options.patch
@@ -1,0 +1,90 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: William Blake Galbreath <blake.galbreath@gmail.com>
+Date: Fri, 25 Jun 2021 19:48:21 -0500
+Subject: [PATCH] Shulker spawn from bullet options
+
+(0 - 1) / 5.0 = -0.2  (can never happen because self is included in count)
+(1 - 1) / 5.0 = 0.0    1.0 - 0.0 = 1.0    100% (just self)
+(2 - 1) / 5.0 = 0.2    1.0 - 0.2 = 0.8     80% (1 other shulker)
+(3 - 1) / 5.0 = 0.4    1.0 - 0.4 = 0.6     60% (2 other shulkers)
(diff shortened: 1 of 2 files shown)


emielderckx:
Maybe the amount of nearby shulkers so it can still work when there are some around

BillyGalbreath:
> Maybe the amount of nearby shulkers so it can still work when there are some around

That's exposed in the equation ^_^

The equation result will be subtracted from the base chance. This defaults to 20% for every nearby shulker

```
(nearby - 1) / 5.0
```

As seen here using 1.0 (100%) as a base chance

```
equation = result     base - result = chance for spawning a new shulker
(0 - 1) / 5.0 = -0.2  (can never happen because self is included in count)
(1 - 1) / 5.0 = 0.0    1.0 - 0.0 = 1.0    100% (just self)
(2 - 1) / 5.0 = 0.2    1.0 - 0.2 = 0.8     80% (1 other shulker)
(3 - 1) / 5.0 = 0.4    1.0 - 0.4 = 0.6     60% (2 other shulkers)
(4 - 1) / 5.0 = 0.6    1.0 - 0.6 = 0.4     40% (3 other shulkers)
(5 - 1) / 5.0 = 0.8    1.0 - 0.8 = 0.2     20% (4 other shulkers)
(6 - 1) / 5.0 = 1.0    1.0 - 1.0 = 0.0      0% (5 other shulkers)
(7 - 1) / 5.0 = 1.2    1.0 - 1.2 = -0.2     0% (6 other shulkers)
```

granny (approved):
LGTM

PR #429, opened 2021-06-28
Redirect System.out calls to plugin loggers

BillyGalbreath:
Opening this as a PR in case anyone has ideas for improvement.

We've been wanting to add this for a long time now, and this is basically a direct copy of [Vicarious' plugin logic](https://github.com/Draycia/Sysout/blob/master/src/main/java/net/draycia/sysoutcatcher/SysoutCatcher.java) that works for Java 8+. I heard there was an easier/better way with the newer Java versions.


katekerllenevich (approved):

YouHaveTrouble (approved):
Don't see any way to improve it further tbh, lgtm and seems to be correct

granny (approved):

PR #448, opened 2021-06-29
Fix and add config spawner silktouch 

Euphillya:
Dont drop xp when spawner is destroy with silktouch and add minimal level require enchant for drop spawner


granny (changes requested):
requires a rebase.

While looking at this I'm a bit confused on the point of `minimumSilkTouchSpawnerRequire`. Silk Touch only has one level; why should this be an option?

YouHaveTrouble:
> While looking at this I'm a bit confused on the point of `minimumSilkTouchSpawnerRequire`. Silk Touch only has one level; why should this be an option?

Technically there are multiple silk touch levels, while only the first one is normally available I've seen plugins utilising the additional ones

Euphillya:
> requires a rebase.
> 
> While looking at this I'm a bit confused on the point of `minimumSilkTouchSpawnerRequire`. Silk Touch only has one level; why should this be an option?

Servers like mine use a level 2 silktouch in order to retrieve the spawner as it can be considered a special item. Plugins do it, it would be a shame not to integrate it while the goal of this patch is not to use a plugin that tampers with the NMS to do what it does.

granny:
> > requires a rebase.
> > While looking at this I'm a bit confused on the point of `minimumSilkTouchSpawnerRequire`. Silk Touch only has one level; why should this be an option?
> 
> Servers like mine use a level 2 silktouch in order to retrieve the spawner as it can be considered a special item. Plugins do it, it would be a shame not to integrate it while the goal of this patch is not to use a plugin that tampers with the NMS to do what it does.

Gotcha. sounds good with me

Euphillya:
Rebase to : https://github.com/pl3xgaming/Purpur/pull/454

PR #458, opened 2021-07-01
migrated to tuinity upstream, no tuinity patches squashing anymore

yannicklamprecht:
- currently there is an issue in paperweight 1.1.7 / 1.1.8 So I downgraded the version for the time being. https://github.com/PaperMC/paperweight/issues/49
- removes the need to manually squash all Tuinity patches into patch one.


patches/server/0001-remove-deobf-of-stacktraces-again.patch
```diff
@@ -0,0 +1,55 @@
...
+
+diff --git a/build.gradle.kts b/build.gradle.kts
+index f766183883f009368b79e1201553c184a373e2cf..fd54481ceaea4e5cbc0e99cee303c1bdeafc5919 100644
+--- a/build.gradle.kts
++++ b/build.gradle.kts
+@@ -133,44 +133,6 @@ tasks.shadowJar {
+     transform(ModifiedLog4j2PluginsCacheFileTransformer::class.java)
+ }
+ 
+-// Paper start - include reobf mappings in jar for stacktrace deobfuscation
```
yannicklamprecht, line 15: Can be re-added when Paperweight has a fix for the above mentioned issue.

katekerllenevich:
we sqash because tuinity does not upstream nearly as much as we would like.

BillyGalbreath:
We don't seem to be effected by that bug. We're able to build Purpur just fine, and I just pushed another patch that will even correct the remapping issues forks are having.

PR #510, opened 2021-07-20
Fix Important Issue Crash with Plugin or Datapack Generation

Euphillya:
Fix Paper#6028 submit to Paper but always not accepted


katekerllenevich:
link to paper pr?

Euphillya:
https://github.com/PaperMC/Paper/pull/6219

katekerllenevich (approved):

SXRWahrheit:
Note that this doesn't actually fix the bug, per https://github.com/PaperMC/Paper/pull/6219, and an additional solution needs to be implemented.

Euphillya:
> Notez que cela ne résout pas réellement le bogue, par [PaperMC / Paper # 6219](https://github.com/PaperMC/Paper/pull/6219), et une solution supplémentaire doit être implémentée.

This is the right solution, if the list is empty, how do you want to use it? Impossible, on vanilla it does not bug because all the trees have leaves.

SXRWahrheit:
The correct result is to not cancel the generation. At the very least, the tree should generate without decorators.

Euphillya:
Its only cancels the decoration.

PR #535, opened 2021-08-01
Fix Crash Issue With Custom Structure DataPack

Euphillya:


Euphillya:
The error no longer recurs without this patch, I don't understand, I close this PR because of this.

PR #570, opened 2021-08-09
Added the ability to add combustible items.

Euphillya:


encode42 (approved):

patches/server/0242-Added-the-ability-to-add-combustible-items.patch
```diff
@@ -0,0 +1,71 @@
...
+index 2c64622577fdb6c2a5746471121403b633bf9042..c012c94012e914db0251a8d16e03efd56f8f18b5 100644
+--- a/src/main/java/net/minecraft/world/level/block/entity/AbstractFurnaceBlockEntity.java
++++ b/src/main/java/net/minecraft/world/level/block/entity/AbstractFurnaceBlockEntity.java
+@@ -205,6 +205,18 @@ public abstract class AbstractFurnaceBlockEntity extends BaseContainerBlockEntit
+         // Paper end
+     }
+ 
++    // Purpur start - Add / Remove item fuel.
++    public static void addFuel(ItemStack itemStack, Integer fuelTimes) {
++        Map<Item, Integer> map = Maps.newLinkedHashMap();
```
BillyGalbreath: Shouldn't this be using the existing `cachedBurnDurations` map so we don't lose all the previous fuels?

Euphillya: I can extract the map to the limit to use it everywhere

BillyGalbreath (approved):

PR #573, opened 2021-08-10
Fix crossbows not working with flame enchant

12emin34:
This resolves https://github.com/pl3xgaming/Purpur/discussions/565


granny (approved):
I've asked 12emin34 to merge this option into the unsafe enchants patch and that it be enabled alongside the `settings.enchantment.allow-unsafe-enchants` option. Would like other's feedback, otherwise LGTM.

katekerllenevich:
I think this fits fine in the unsafe enchants patch, LGTM.

PR #643, opened 2021-09-04
configurable damage for aggressive cows

Oharass:
purpur is for configuration. Why does the kernel not have setting for cow damage? This adds.


granny (dismissed):
```yaml
cow:
  naturally-aggressive-to-players:
    chance: ##
    damage: ##
```
requires config migration for the above change ^

BillyGalbreath (approved):

granny (approved):

PR #644, opened 2021-09-04
Remove references to Tuinity

fredster33:
I am 99.99 percent sure that I did something I wasn't supposed to. I'm very unfamiliar with the patch system, so if I updated something upstream or something that I wasn't supposed to, please tell me.


BillyGalbreath:
Unfortunately, this will make the builds fail. We aren't the ones adding the Tuinity references in most of these, airplane is

katekerllenevich:
Yeah pretty much all of these are a no, lol. The changes you made (except for the last) are from the Airplane patch. The Airplane patch is completely identical to the patches from [Airplane](https://github.com/TECHNOVE/Airplane/), and the last one is legitimate.

fredster33:
Got it.

PR #657, opened 2021-09-12
permission for idle kick

Oharass:
https://github.com/pl3xgaming/Purpur/discussions/656

This adds

### DIFF
--- a/patches/server/0013-AFK-API.patch
+++ b/patches/server/0013-AFK-API.patch
@@ -68,23 +68,44 @@ index d2d2e3c58bd3d08f4e7f6257502ce8c2bca19825..04d2e4e641c12fb259ee7758db81b354
          return this.stats;
      }
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
-index 88227a0882b4a667206891b2a26ad5764fd33590..1ba2155e353b390cff90605f3158b8f61d531cb3 100644
+index 88227a0882b4a667206891b2a26ad5764fd33590..ecd0fa7daa11cd9f27b2e48cf935a5e92ee5b2f4 100644
 --- a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
 +++ b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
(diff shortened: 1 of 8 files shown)


granny (changes requested):
needs to be merged into `0013-AFK-API.patch`

Oharass:
ok.

Oharass:
done

katekerllenevich (approved):

patches/server/0013-AFK-API.patch
```diff
@@ -77,7 +77,7 @@ index eaca5e3c6d26d5fff1eae1cacabfd389b74aec4d..9de4ca41c22fc92762b8c2705cdc8a5f
          if (this.player.getLastActionTime() > 0L && this.server.getPlayerIdleTimeout() > 0 && Util.getMillis() - this.player.getLastActionTime() > (long) (this.server.getPlayerIdleTimeout() * 1000 * 60) && !this.player.wonGame) { // Paper - Prevent AFK kick while watching end credits.
 +            // Purpur start
 +            this.player.setAfk(true);
-+            if (!this.player.level.purpurConfig.idleTimeoutKick) {
++            if (!this.player.level.purpurConfig.idleTimeoutKick || org.bukkit.Bukkit.getPlayer(this.player.getUUID()).hasPermission("purpur.bypassIdleKick")) {
```
BillyGalbreath: we need to find a way to cache this permission check so it's not being performed every single tick

BillyGalbreath (changes requested):

patches/server/0013-AFK-API.patch
Oharass: I will.

YouHaveTrouble: why not just check that every 5-10 ticks?
Shouldn't make a big difference afaik

Euphillya: And we can also use : `this.player.getBukkitEntity().hasPermission("purpur.bypassIdleKick")`

Oharass: I had no idea such a method existed!!! Perhaps there should be javadoc published for this. It will make development much easier. I will evaluate the best way to reduce permission check when done working.

Oharass: it caches. I add. Tell me discord Oharass#8311 on PurPur discord if I do wrong or here.

PR #661, opened 2021-09-14
Add water damage to all mobs

sph432:


patches/server/0259-add-water-damage-to-all-endermobs-and-magmacube.patch
```diff
@@ -0,0 +1,106 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+index dd9655f29c397d3df15685f2901e1f232f17fec1..0d6691484773c97bed2965f48d5494bff34f2403 100644
+--- a/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
++++ b/src/main/java/net/minecraft/world/entity/boss/enderdragon/EnderDragon.java
+@@ -161,6 +161,11 @@ public class EnderDragon extends Mob implements Enemy {
+     }
+     // Purpur end
+ 
++    @Override
++    public boolean isSensitiveToWater() {
```
BillyGalbreath: Move this whole method above `// Purpur end` since we're adding it to this class

patches/server/0259-add-water-damage-to-all-endermobs-and-magmacube.patch
```diff
@@ -0,0 +1,106 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/monster/Endermite.java b/src/main/java/net/minecraft/world/entity/monster/Endermite.java
+index 1578a8196f860995370a1b8938b6d71040f5d3a9..7c692c99e27823a506bde48402b559fe5a0c9b6a 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/Endermite.java
++++ b/src/main/java/net/minecraft/world/entity/monster/Endermite.java
+@@ -62,6 +62,11 @@ public class Endermite extends Monster {
+     }
+     // Purpur end
+ 
++    @Override
++    public boolean isSensitiveToWater() {
```
BillyGalbreath: Move this whole method above `// Purpur end` since we're adding it to this class

patches/server/0259-add-water-damage-to-all-endermobs-and-magmacube.patch
```diff
@@ -0,0 +1,106 @@
...
+         this.goalSelector.addGoal(1, new FloatGoal(this));
+diff --git a/src/main/java/net/minecraft/world/entity/monster/MagmaCube.java b/src/main/java/net/minecraft/world/entity/monster/MagmaCube.java
+index 87aae8fa18704e58f2b1fc0abd1a98987715123f..ed8beeaf6190e20b589935245fa7a5326015c5ef 100644
+--- a/src/main/java/net/minecraft/world/entity/monster/MagmaCube.java
++++ b/src/main/java/net/minecraft/world/entity/monster/MagmaCube.java
+@@ -54,6 +54,11 @@ public class MagmaCube extends Slime {
+     }
+     // Purpur end
+ 
++    @Override
```
BillyGalbreath: Move this whole method above `// Purpur end` since we're adding it to this class

BillyGalbreath (changes requested):

BillyGalbreath:
I like it. Maybe we should extend this to all mobs?

YouHaveTrouble:
> I like it. Maybe we should extend this to all mobs?

Yes. 

Also, shouldn't it be in [this patch](https://github.com/pl3xgaming/Purpur/blob/ver/1.17.1/patches/server/0153-Toggle-for-water-sensitive-mob-damage.patch)?

granny:
> > I like it. Maybe we should extend this to all mobs?
> 
> Yes.
> 
> Also, shouldn't it be in [this patch](https://github.com/pl3xgaming/Purpur/blob/ver/1.17.1/patches/server/0153-Toggle-for-water-sensitive-mob-damage.patch)?

merging with the mentioned patch would be preferred

sph432:
Will change that patch then, I did notice the patch, but did not know about the etiquette of changing another's patch so better be safe, I submitted another.

sph432:
extended config options to all mobs
moved changes back to original patch
fixed formatting issues

granny (changes requested):
the naming of some config options does not match

`mobs.caveSpider.takes-damage-from-water`
should be
`mobs.cave_spider.takes-damage-from-water`
etc.

also requires a rebase

PR #685, opened 2021-09-26
fix empty messages still sending when empty

### LINKED ISSUE #684: Empty value cannot-ride-mob sends empty chat message
### Timings link

_No response_

### Expected behavior

Setting the value `cannot-ride-mob` to an empty string should not send a chat message when applicable.

### Observed/Actual behavior

Setting the value `cannot-ride-mob` to an empty string actually sends an empty chat message when applicable.

### Steps/models to reproduce

`cannot-ride-mob: ''` in config 

### Purpur version

This server is running Purpur version git-Purpur-1394 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: acb8a26 on ver/1.17.1)
You are running the latest version
Previous version: git-Purpur-1375 (MC: 1.17.1)

### Agreements

- [X] I am running the latest version of Purpur available from https://purpur.pl3x.net/downloads.

### Other

Issue #53 suggests this was previously possible.

Oharass:
Fix #684. Use empty message for cannot-ride-mob. This adds

### DIFF
--- a/patches/server/0006-Component-related-conveniences.patch
+++ b/patches/server/0006-Component-related-conveniences.patch
@@ -103,7 +103,7 @@ index ad9fb50791779a5fe7d22268b71bd10d9c9ff3f0..1c1602f839828252748e1acacfe42c7c
          this.server.sendMessage(message, sender);
          Iterator iterator = this.players.iterator();
 diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
-index 732eb171f6a8fe1b02044a00fcf85217c41116ac..3b7d72f035d7fd74971af043294047ff1ed4c95c 100644
+index 5d455a662107182659abcbe83195ac5d6cf25813..22a730d1b2f8ede06ae3366344af5120134d1af7 100644
 --- a/src/main/java/net/minecraft/world/entity/Entity.java
 +++ b/src/main/java/net/minecraft/world/entity/Entity.java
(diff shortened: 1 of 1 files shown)


katekerllenevich (approved):

BillyGalbreath (changes requested):
While this would fix it for this specific option, it won't for any of the others. A better way is to check in the `sendMessage` method. Seems like when I switched everything to components that logic was lost somewhere.

Oharass:
ok. I will fix this.

Oharass:
I think this is correct now. Fixed for all. I use "isEmpty" instead of "isBlank" so you can still send blank message if you like with " ". This adds.

patches/server/0006-Component-related-conveniences.patch
```diff
@@ -116,7 +116,7 @@ index 732eb171f6a8fe1b02044a00fcf85217c41116ac..3b7d72f035d7fd74971af043294047ff
 +    }
 +
 +    public void sendMessage(@Nullable String message, UUID sender) {
-+        if (message != null) {
++        if (org.apache.commons.lang3.StringUtils.isNotEmpty(message)) {
```
encode42, line 119: Why use StringUtils instead of Java's native `String#isEmpty`?

Oharass: It handles null. I may also do `message != null && !message.isEmpty()`

encode42: Ah, I see. Was just wondering

encode42 (changes requested):
LGTM, just resolve conflicts

PR #686, opened 2021-09-26
add new event

Oharass:
add event for player packets. Packet can not be api ever because too super unstable. Because of this, packet is object. Nothing more nothing less. Not real packet. Just object. This is useful for plugin like protocollib can not do easy. This is most expansive move for purpur api yet. This adds.

if you do not like tell me. I accept defeat.

This is much more useful because of MOJMAP and PAPERWEIGHT. Much more stable for different versions.

I would like to add a send packet method in the future with this. This new method can replace protocollib without true packet api.

### DIFF
--- a/patches/api/0051-add-player-packet-event.patch
+++ b/patches/api/0051-add-player-packet-event.patch
@@ -1,0 +1,68 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: MiniDigger <admin@minidigger.me>
+Date: Sat, 25 Sep 2021 21:34:34 -0500
+Subject: [PATCH] add player packet event
+
+Co-authored by: oharass <oharass@bk.ru>
+
+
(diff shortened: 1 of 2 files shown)


BomBardyGamer:
Hope no one complains about me sharing my view on this, but I just want to say something here.

The reason why nobody has done this before is because, as you said yourself, a packet API is too unstable. Also, it is way too low-level to implement in the way that most would be seeking.
Also, I believe that having packets here being `Object`s doesn't really help either, since it:
1. Promotes the usage of NMS, which should be avoided under all circumstances
2. Makes it clear that this should not be public API

In summary, it's a good addition, just not one to be exposed to public API.

(P.S. I realise I'm really late here, but I thought I should voice my thoughts anyway, because why not)

katekerllenevich:
I'm going to deny this PR for a few reasons.
1) There are other ways to register a listener for packets (https://www.spigotmc.org/threads/packet-listener-using-netty-tutorial.69772/)
2) Bukkit events are laggy AF, and packets are sent *all* the time
3) Packets never ever touch API

PR #687, opened 2021-09-26
fix offline mode message

Oharass:
Purpur has a FLAWED check for if bungee proxy online is in use. isProxyOnlineMode is TRUE BY DEFAULT. Even without bungee enabled. This is big issue. It does not support velocity either. This adds.

### DIFF
--- a/patches/server/0003-Rebrand.patch
+++ b/patches/server/0003-Rebrand.patch
@@ -169,7 +169,7 @@ index f6d335b377dbf6f744e5f6ced56cf437e6e6bab3..306b4185fcc86189966f1170547ca1d6
  
      public SystemReport fillSystemReport(SystemReport details) {
 diff --git a/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java b/src/main/java/net/minecraft/server/dedicated/DedicatedServer.java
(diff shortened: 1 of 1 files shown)


katekerllenevich:
isProxyOnlineMode checks org.spigotmc.SpigotConfig.bungee unless i'm mistaken

Euphillya:
Yes it checks well:
https://github.com/PaperMC/Paper/blob/06db5d08b08c4f53fea9cb322127b3a3311bbcd9/patches/server/0103-Add-setting-for-proxy-online-mode-status.patch#L30

Oharass:
Yes. This supports velocity. The message contains velocity but velocity is not supported.

```java
                DedicatedServer.LOGGER.warn("Whilst this makes it possible to use BungeeCord or Velocity, unless access to your server is properly restricted, it also opens up the ability for hackers to connect with any username they choose."); // Purpur
                DedicatedServer.LOGGER.warn("Please see http://www.spigotmc.org/wiki/firewall-guide/ for further information.");
```

Euphillya:
https://github.com/PaperMC/Paper/blob/cc6fd371b2d1ab2ad1df85789b143f04a17c1468/patches/server/0280-Add-Velocity-IP-Forwarding-Support.patch#L33

Velocity supported

PR #688, opened 2021-09-26
auto offline mode bungee velocity

Oharass:
Many users are afraid of offline mode but need to use bungee and support. This is because of a lack of understanding. In addition to #687; this will make it easier for bungee and velocity users without additional configuration. To make it easier. This adds.


Euphillya:
This PR has already been offered and refused : #340

Oharass:
I did not see. Ok.

PR #698, opened 2021-10-04
Configurable Max TPS option

Oharass:
Purpur is about creating unique experiences for server owners and no plugin may replicate this. You can set and reload the TPS of the server. In future, I may add command if it is desired feature. I can even add "tick forward/warp" command like CARPET mod. https://www.youtube.com/watch?v=_Me-Ik_20Gg. This adds.

Config loads too late to set TPS properly, so I add new method. Tell me if this is wrong. I must also manually reload. I may also make config early bird but this seems bad plan. This adds.


Euphillya:
It is not likely to crash plugins? Isn't it easier just to change the randomTickSpeed ​​gamerule?

granny:
After some discussion in the discord server, the decision has been made to close this PR.

PR #709, opened 2021-10-15
axolotl options to choose hunting/hostile targets

granny:
Whenever you run the `/purpur reload`, all axolotls will "freeze" until the command finishes running (which makes sense since targets are tied to the config options now).

Attempting to remove a mob from the list and running `/purpur reload` will not actually remove the mob from the list; it will just add onto what the list already has (so essentially duplicates the list if there's no changes). I replicated how `getList` is used similarly to other `getList` declarations in `PurpurWorldConfig` soo not sure if this is an issue with just my usage of `getList` or if it's reproducible with the other list options as well.

### DIFF
--- a/patches/server/0261-axolotl-options-to-choose-hunting-hostile-targets.patch
+++ b/patches/server/0261-axolotl-options-to-choose-hunting-hostile-targets.patch
@@ -1,0 +1,70 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: granny <granny@pl3x.net>
+Date: Fri, 15 Oct 2021 02:22:19 -0700
+Subject: [PATCH] axolotl options to choose hunting/hostile targets
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java b/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java
+index 4433c567cce113dc7cd53b517beb94f63ed2606c..cc5de6bcaad6bfc0cfc1292de272aad1b89c782b 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java
++++ b/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java
+@@ -17,11 +17,11 @@ public class AxolotlAttackablesSensor extends NearestVisibleLivingEntitySensor {
+     }
+ 
+     private boolean isHuntTarget(LivingEntity axolotl, LivingEntity target) {
+-        return !axolotl.getBrain().hasMemoryValue(MemoryModuleType.HAS_HUNTING_COOLDOWN) && EntityTypeTags.AXOLOTL_HUNT_TARGETS.contains(target.getType());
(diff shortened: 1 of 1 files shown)


patches/server/0262-axolotl-options-to-choose-hunting-hostile-targets.patch
```diff
@@ -0,0 +1,77 @@
...
+
+
+diff --git a/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java b/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java
+index 4433c567cce113dc7cd53b517beb94f63ed2606c..f1703bd9925353e22438b4a08f99049ca680a570 100644
+--- a/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java
++++ b/src/main/java/net/minecraft/world/entity/ai/sensing/AxolotlAttackablesSensor.java
+@@ -1,6 +1,5 @@
+ package net.minecraft.world.entity.ai.sensing;
+ 
+-import net.minecraft.tags.EntityTypeTags;
```
katekerllenevich: unnecessary diff

katekerllenevich (approved):

patches/server/0261-axolotl-options-to-choose-hunting-hostile-targets.patch
```diff
@@ -0,0 +1,70 @@
...
+ 
+     private boolean isHuntTarget(LivingEntity axolotl, LivingEntity target) {
+-        return !axolotl.getBrain().hasMemoryValue(MemoryModuleType.HAS_HUNTING_COOLDOWN) && EntityTypeTags.AXOLOTL_HUNT_TARGETS.contains(target.getType());
++        return !axolotl.getBrain().hasMemoryValue(MemoryModuleType.HAS_HUNTING_COOLDOWN) && axolotl.level.purpurConfig.axolotlHuntTargets.contains(target.getType()); // Purpur
+     }
+ 
+     private boolean isHostileTarget(LivingEntity axolotl) {
+-        return EntityTypeTags.AXOLOTL_ALWAYS_HOSTILES.contains(axolotl.getType());
++        return axolotl.level.purpurConfig.axolotlAlwaysHostiles.contains(axolotl.getType()); // Purpur
+     }
```
granny, line 22: Need to look into not using ArrayList's contains method

granny:
Initially when I made this PR I had issues with modifying the axolotl entity type tags through a datapack, so I created this patch to make it possible through Purpur. I decided to come back to this and see if I could modify the `axolotl_hunt_targets` and `axolotl_always_hostiles` entity type tags through a datapack, and it worked this time (I'm not sure what I did wrong previously). I don't see the point in rebasing/managing/improving upon this patch when it can be handled through a datapack. I'm sure there are a few people who are interested in how to make this possible through a datapack, so I've outlined some useful information below.

A direct link to the current default values for each entity type tag.
https://minecraft.fandom.com/wiki/Tag#Entity_types

The paths & contents of the json files as of `1.18.2`.

`/data/minecraft/tags/entity_types/axolotl_hunt_targets.json`
```json
{
  "replace": true,
  "values": [
    "minecraft:tropical_fish",
    "minecraft:pufferfish",
    "minecraft:salmon",
    "minecraft:cod",
    "minecraft:squid",
    "minecraft:glow_squid"
  ]
}
```

`/data/minecraft/tags/entity_types/axolotl_always_hostiles.json`
```json
{
  "replace": true,
  "values": [
    "minecraft:drowned",
    "minecraft:guardian",
    "minecraft:elder_guardian"
  ]
}
```

PR #783, opened 2021-12-12
Add Potion NamespacedKey Patches

DaRacci:
New PR into 1.18.1, no changes since last PR

### DIFF
--- a/patches/api/0043-Potion-NamespacedKey.patch
+++ b/patches/api/0043-Potion-NamespacedKey.patch
@@ -1,0 +1,229 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Racci <tangentmoons@gmail.com>
+Date: Fri, 3 Dec 2021 23:48:26 +1100
+Subject: [PATCH] Potion NamespacedKey
+
+
+diff --git a/src/main/java/org/bukkit/potion/PotionEffect.java b/src/main/java/org/bukkit/potion/PotionEffect.java
+index 74767751199bce03d63f2a9524712656193f850c..daa3365d48d2d3e3763f5f4440f3afbc7982c95e 100644
+--- a/src/main/java/org/bukkit/potion/PotionEffect.java
++++ b/src/main/java/org/bukkit/potion/PotionEffect.java
+@@ -5,6 +5,7 @@ import java.util.Map;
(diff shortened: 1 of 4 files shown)


encode42 (approved):

Camotoy:
I don't understand the purpose of this PR. Doesn't the persistent data container exist for this reason? Is there something here that existing functionality doesn't provide? Besides that, have you tested if this data stays should the user switch from Purpur to another downstream server build? Also, why specifically for this one entity type? In my opinion, this seems like a really, really niche feature.

Camotoy:
My apologies; this is for actual effects, not any sort of entity. Though the point about the data persisting between server forks *should* still persist.

DaRacci:
> My apologies; this is for actual effects, not any sort of entity. Though the point about the data persisting between server forks _should_ still persist.

I haven't done any form of testing between forks however due to modifying the serializer the data wont persist through forks as it won't load the key and would override it once saved again

patches/api/0043-Potion-NamespacedKey.patch
```diff
@@ -0,0 +1,229 @@
...
+@@ -242,9 +298,9 @@ public class PotionEffect implements ConfigurationSerializable {
+      * @return color of this potion's particles. May be null if the potion has no particles or defined color.
+      * @deprecated color is not part of potion effects
+      */
+-     @Deprecated
+-     @Nullable
+-     @Contract("-> null")
++    @Deprecated
++    @Nullable
++    @Contract("-> null")
```
encode42: Unmodify these lines so we don't run into upstream conflicts

encode42 (changes requested):

encode42:
> Though the point about the data persisting between server forks should still persist.

Due to the nature of NBT's implementation, all unused data will be wiped. This applies to Paper and all other Purpur patches that modify NBT data. The general census is that when you switch forks, you'll lose all of the fork's features.

Because of this, I believe it's fine if the data does not persist between forks.

patches/api/0043-Potion-NamespacedKey.patch
DaRacci: I cannot count how many times ive fixed this now, For some reason intellij keeps formatting them and i dont notice until i get told about it like this

BillyGalbreath (approved):

PR #792, opened 2021-12-16
remove useless patch

Oharass:
Paper fix this on October 3. https://github.com/PaperMC/Paper/pull/6452. Purpur patch no longer required.


BillyGalbreath:
I'm not sure I'm ok with their fix there, tbh. We've had an issue before with this method and we fixed it in a different way. It was a mappings issue before, and I have a hunch it's going to be another mappings issue here again. Let me look deeper into this before deciding how to proceed.

Oharass:
paper fix since then https://github.com/PaperMC/paperweight/blob/main/paperweight-lib/src/main/kotlin/tasks/FixJarForReobf.kt#L132. But I know caution

BillyGalbreath:
If it's the same as before, then it's not something Paper could fix. It was a mappings issue in CraftBukkit. https://github.com/PaperMC/Paper/issues/5994#issuecomment-868950796

PR #809, opened 2021-12-28
more spawner settings

Oharass:
This is very frequently requested setting especially in paper discord, for people who want spawner globally slower or faster or other spawner change!

NOTE

The `getOrCreateDisplayEntity` method is not used. But I put there anyway in the case that it is used in future. If it is, change to `getUpdateTag` method not required!.


BillyGalbreath:
The reason I havent done this is because it doesnt work for existing spawners. The values are saved to NBT, and so far all attempts at making them work for existing spawners have conflicted with API usage.

Oharass:
Yes I understand, this is big limitation. I will try to 1. fix or if I can not I will close.

Oharass:
I will not fix this soon. I may reopen in the future but I have no time now.

PR #837, opened 2022-01-07
Configurable minimum demand for trades

totorewa:
Paper adds [a patch](https://github.com/PaperMC/Paper/blob/master/patches/server/0436-Fix-villager-trading-demand-MC-163962.patch) which stops demand for villager trade offers from going below zero, addressing the issue tracked by [MC-163962](https://bugs.mojang.com/browse/MC-163962). This does have a noticeable effect on trading, as prices can increase quite substantially (with cure discounts) after a single restock where as in vanilla it can take a few restocks to happen.  
  
It's unknown whether going below zero is intentional. It could be but maybe not quite as far in to the negatives as it does go if you don't trade for a long time.

This patch extends the change made by the previous patch by making the minimum demand configurable to leave it up to server administrators to decide.

### DIFF
--- a/patches/server/0257-Configurable-minimum-demand-for-trades.patch
+++ b/patches/server/0257-Configurable-minimum-demand-for-trades.patch
@@ -1,0 +1,64 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Totorewa <76272501+totorewa@users.noreply.github.com>
+Date: Fri, 7 Jan 2022 21:34:57 +1300
+Subject: [PATCH] Configurable minimum demand for trades
+
+Addresses MC-163962 where villager demand decreases indefinitely. Paper
+adds a patch to fix this by preventing demand from going below zero.
+This patch adds a config option to allow the minimum demand to instead
+be configurable.
+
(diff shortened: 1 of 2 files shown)


patches/server/0257-Configurable-minimum-demand-for-trades.patch
```diff
@@ -0,0 +1,64 @@
...
+diff --git a/src/main/java/net/minecraft/world/item/trading/MerchantOffer.java b/src/main/java/net/minecraft/world/item/trading/MerchantOffer.java
+index c9cb0717c2793acd5b5870a6cc4d672d69a40026..1b4ec1e2f25c66f9372aa004224556a9a7307ae1 100644
+--- a/src/main/java/net/minecraft/world/item/trading/MerchantOffer.java
++++ b/src/main/java/net/minecraft/world/item/trading/MerchantOffer.java
+@@ -132,9 +132,15 @@ public class MerchantOffer {
+     }
+ 
+     public void updateDemand() {
+-        this.demand = Math.max(0, this.demand + this.uses - (this.maxUses - this.uses)); // Paper
++        this.updateDemand(0); // Purpur
```
BillyGalbreath: Why `0` here instead of `villagerMinimumDemand`?

BillyGalbreath: Nevermind. I see it now. No instance to `level` in this context.

PR #850, opened 2022-01-16
Additional configuration for grindstones

encode42:
Adds a list of enchants the grindstone should ignore when unenchanting items. This *shouldn't* affect performance in any way, but if anyone knows different, lmk. Useful for custom curses, ignoring nerf enchants, etc.

Also adds the ability to remove attributes, such as off-hand speed, jump height, etc.

### DIFF
--- a/patches/server/0253-Shears-can-have-looting-enchantment.patch
+++ b/patches/server/0253-Shears-can-have-looting-enchantment.patch
@@ -158,10 +158,10 @@ index 6b8a1535086aae7e4e3229d05615fb903188f507..60af917083de1b790b1d93d61835a669
      public int getMinCost(int level) {
          return 15 + (level - 1) * 9;
 diff --git a/src/main/java/org/purpurmc/purpur/PurpurConfig.java b/src/main/java/org/purpurmc/purpur/PurpurConfig.java
-index 6eb138a83460e049f711c67055fe6db4ff812e8c..f19e6c9af8cd559ba5a1492b0d579bf6182f342a 100644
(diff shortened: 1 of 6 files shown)


patches/server/0253-Shears-can-have-looting-enchantment.patch
```diff
@@ -158,18 +158,18 @@ index 6b8a1535086aae7e4e3229d05615fb903188f507..60af917083de1b790b1d93d61835a669
...
-index 6eb138a83460e049f711c67055fe6db4ff812e8c..f19e6c9af8cd559ba5a1492b0d579bf6182f342a 100644
+index 7d3f8eb6db6193b8151d732232171a9a455e1470..469f40bf9a31a254950e56b85605c5e729f22e16 100644
 --- a/src/main/java/org/purpurmc/purpur/PurpurConfig.java
 +++ b/src/main/java/org/purpurmc/purpur/PurpurConfig.java
-@@ -340,6 +340,7 @@ public class PurpurConfig {
+@@ -357,6 +357,7 @@ public class PurpurConfig {
  
      public static boolean allowInfinityMending = false;
      public static boolean allowCrossbowInfinity = false;
 +    public static boolean allowShearsLooting = false;
```
YouHaveTrouble, line 168: how does that relate to the grindstone patch?

katekerllenevich: that has nothing to do with it, note that the line in question isn't being modified. its just part of a patch file that was

PR #866, opened 2022-01-24
More Paper -> Purpur change

TetraTheta:
This is very small change.

All it does is just changing Paper download link to Purpur download link in message which is shown when running server with very outdated build of server.


BillyGalbreath:
This build check is actually completely bypassed on Purpur.

https://github.com/PurpurMC/Purpur/blob/ver/1.18.1/patches/server/0026-Disable-outdated-build-check.patch

PR #891, opened 2022-02-04
Add Fire ImmunityAPI

DaRacci:


patches/api/0048-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,54 @@
...
+-     */
+-    void setImmuneToFire(boolean immuneToFire);
+-
+-    /**
+-     * Check if item is immune to fire
+-     *
+-     * @return True if immune to fire
+-     */
+-    boolean isImmuneToFire();
+-
```
BillyGalbreath: I'll go over the rest of this PR when I have more time, but just wanted to point out that this breaks [ABI](https://stackoverflow.com/questions/2171177/what-is-an-application-binary-interface-abi). We'll have to keep these methods here for compatibility, and in the implementation just send them to `super`

BillyGalbreath (changes requested):

patches/api/0048-Fire-Immunity-API.patch
DaRacci: Thank for pointing to that info on ABI as i didn't know about that, and on the point of this since in CraftItem it needs to override both the nullable boolean and primitive would the best way of doing this be to be adding in this due to needing to call this on the item which is what the other override calls
```java
@Override
public void setImmuneToFire(boolean immuneToFire) {
    this.setImmuneToFire((Boolean) immuneToFire);
}
```

BillyGalbreath: I didn't even notice the difference in method params. That'll work, since the method signatures are different. But, I prefer the super way because it's more clear what it's doing and doesn't need an explicit cast. ;)
```java
    @Override
    public void setImmuneToFire(boolean immuneToFire) {
        super.setImmuneToFire(immuneToFire);
    }
```

However, this raises another concern, in that there will be 2 distinct methods for this now. So we might want to deprecate this one in favor of the new one in Entity as well.

image

DaRacci: When i was doing my initial testing using super in the CraftItem didn't work since the entity which needs to be effects is the item field from what i saw if i didn't somehow mess it up, and I'll add a deprecated annotation onto the old method too

patches/server/0269-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,107 @@
...
+
+diff --git a/src/main/java/net/minecraft/world/entity/Entity.java b/src/main/java/net/minecraft/world/entity/Entity.java
+index 44e693e88e03cd34ac0fd293e611a0ebf1221f53..57bd455d3cae42aac8f24b6cf6ead4ffca418604 100644
+--- a/src/main/java/net/minecraft/world/entity/Entity.java
++++ b/src/main/java/net/minecraft/world/entity/Entity.java
+@@ -92,7 +92,6 @@ import net.minecraft.world.level.block.Mirror;
+ import net.minecraft.world.level.block.RenderShape;
+ import net.minecraft.world.level.block.Rotation;
+ import net.minecraft.world.level.block.SoundType;
+-import net.minecraft.world.level.block.state.BlockBehaviour;
```
katekerllenevich: don't mess with this

patches/server/0269-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,107 @@
...
+ }
+diff --git a/src/main/java/org/bukkit/craftbukkit/entity/CraftItem.java b/src/main/java/org/bukkit/craftbukkit/entity/CraftItem.java
+index 0637e074dbe4a10bc19e7bf9dd79e2d966b07b1d..a13b2da84477f3e58a01150008a6b7109710b063 100644
+--- a/src/main/java/org/bukkit/craftbukkit/entity/CraftItem.java
++++ b/src/main/java/org/bukkit/craftbukkit/entity/CraftItem.java
+@@ -8,6 +8,7 @@ import org.bukkit.craftbukkit.inventory.CraftItemStack;
+ import org.bukkit.entity.EntityType;
+ import org.bukkit.entity.Item;
+ import org.bukkit.inventory.ItemStack;
++import org.jetbrains.annotations.*;
```
katekerllenevich: use fqdn

patches/server/0269-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,107 @@
...
+ 
+ public class CraftItem extends CraftEntity implements Item {
+ 
+@@ -153,9 +154,14 @@ public class CraftItem extends CraftEntity implements Item {
+         return item.immuneToExplosion;
+     }
+ 
++    @Override
++    public void setImmuneToFire(@Nullable Boolean immuneToFire) {
```
katekerllenevich: `boolean` not `Boolean`

patches/server/0269-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,107 @@
...
+ import net.minecraft.world.level.block.SoundType;
+-import net.minecraft.world.level.block.state.BlockBehaviour;
+ import net.minecraft.world.level.block.state.BlockState;
+ import net.minecraft.world.level.block.state.properties.BlockStateProperties;
+ import net.minecraft.world.level.border.WorldBorder;
+@@ -334,6 +333,7 @@ public abstract class Entity implements Nameable, EntityAccess, CommandSource, i
+     @javax.annotation.Nullable
+     private UUID originWorld;
+     public boolean freezeLocked = false; // Paper - Freeze Tick Lock API
++    public @Nullable Boolean immuneToFire = null; // Purpur - Fire immune API
```
katekerllenevich: `boolean` not `Boolean`

patches/api/0048-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,54 @@
...
++     *
++     * @return True if fire immune
++     */
++    boolean isImmuneToFire();
++
++    /**
++     * Sets if the entity is fire immune
++     * Set this to null to restore the entity type default
++     */
++    void setImmuneToFire(@Nullable Boolean fireImmune);
```
katekerllenevich: `boolean` not `Boolean`

patches/api/0048-Fire-Immunity-API.patch
```diff
@@ -0,0 +1,54 @@
...
+ import org.bukkit.inventory.ItemStack;
+ import org.jetbrains.annotations.NotNull;
+ import org.jetbrains.annotations.Nullable;
+@@ -173,7 +177,9 @@ public interface Item extends Entity {
+      * Set whether or not this item is immune to fire
+      *
+      * @param immuneToFire True to make immune to fire
++     @deprecated use {@link #setImmuneToFire(Boolean)} instead
+      */
++    @Deprecated
```
katekerllenevich: i'm not sure this is necessary

katekerllenevich (changes requested):

patches/server/0269-Fire-Immunity-API.patch
DaRacci: the idea of having it be Boolean is so it can be set to null and restore default behaviour which i think is a better way than using another method or something to signal that its disabled

katekerllenevich: Maybe take a look at how I did this?
https://github.com/PurpurMC/Purpur/blob/ver/1.18.2/patches/api/0035-API-for-any-mob-to-burn-daylight.patch

DaRacci: I think its better to be able to do it this way because if you want to set the behaviour to the default for the mob you'd have to know the mob and its default setting, however with this method and how I've been using it in my development has been used with a system that doesn't track a certain mob and doesn't know the default so being able to null it and have it just work makes it so much easier to use and work with.

DaRacci: unless im missing something the only way to get the default fire immune setting is like this
```kotlin
val default = (entity as CraftEntity).handle.type.fireImmune()
entity.setImmuneToFire(default)
```
but with being able to null it
```kotlin
entity.setImmuneToFire(null)
```
This would mean to be able to reset to the default you either need to know the mob type and its default or use nms.

patches/api/0048-Fire-Immunity-API.patch
DaRacci: While it may not be necessary i think deprecating this and pointing towards the universal fire immune method is a better idea

patches/server/0269-Fire-Immunity-API.patch
4drian3d: Then it may be better to use a [TriState](https://jd.adventure.kyori.net/api/4.10.1/net/kyori/adventure/util/TriState.html) instead of a Boolean which may be null. This way you can avoid using booleans as objects and avoid having a null variable

katekerllenevich: I still think thats unnecessary and that a simple `boolean` (matching with the burn in daylight API) is fine.

granny (approved):

PR #940, opened 2022-03-24
Configurable map update interval

minhh2792:
Mojang decided to increase the update interval of map to 5 ticks, in some server which utilize it this is kinda slow

This PR made the value configurable so the server admin can adjust as their liking

I tested this and having no issue, however I can't record a comparison video of it because of my potato pc :'(

### DIFF
--- a/patches/server/0272-Configurable-map-update-interval.patch
+++ b/patches/server/0272-Configurable-map-update-interval.patch
@@ -1,0 +1,44 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Minh Nguyen <minhh2792.work@gmail.com>
+Date: Thu, 24 Mar 2022 04:27:22 +0000
+Subject: [PATCH] Configurable map update interval
+
+
+diff --git a/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java b/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
+index 6a78374b6ab3198571df7e17023d371667ebb24b..19d9fe2302a7acb9f4aec0e80ce81354248ed0f0 100644
+--- a/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
++++ b/src/main/java/net/minecraft/world/level/saveddata/maps/MapItemSavedData.java
+@@ -624,7 +624,7 @@ public class MapItemSavedData extends SavedData {
+         @Nullable
+         Packet<?> nextUpdatePacket(int mapId) {
+             MapItemSavedData.MapPatch worldmap_b;
+-            if (!this.dirtyData && this.tick % 5 != 0) { this.tick++; return null; } // Paper - this won't end up sending, so don't render it!
++            if (!this.dirtyData && this.tick % org.purpurmc.purpur.PurpurConfig.mapUpdateInterval != 0) { this.tick++; return null; } // Paper - this won't end up sending, so don't render it! // Purpur - configurable map update interval
+             boolean vanillaMaps = shouldUseVanillaMap(); // Paper
+             org.bukkit.craftbukkit.map.RenderData render = !vanillaMaps ? MapItemSavedData.this.mapView.render((org.bukkit.craftbukkit.entity.CraftPlayer) this.player.getBukkitEntity()) : MapItemSavedData.this.vanillaRender; // CraftBukkit // Paper
+ 
+@@ -637,7 +637,7 @@ public class MapItemSavedData extends SavedData {
+ 
+             Collection collection;
+ 
+-            if ((true || this.dirtyDecorations) && this.tick++ % 5 == 0) { // CraftBukkit - custom maps don't update this yet
++            if ((true || this.dirtyDecorations) && this.tick++ % org.purpurmc.purpur.PurpurConfig.mapUpdateInterval == 0) { // CraftBukkit - custom maps don't update this yet // Purpur - configurable map update interval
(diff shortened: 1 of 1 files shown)


patches/server/0272-Configurable-map-update-interval.patch
```diff
@@ -0,0 +1,55 @@
...
+index 659e93fb24124318367e5b4d0b144ca1443950e9..6a37582f563dd3eaf16a9bb434490decd03b25dc 100644
+--- a/src/main/java/org/purpurmc/purpur/PurpurConfig.java
++++ b/src/main/java/org/purpurmc/purpur/PurpurConfig.java
+@@ -72,8 +72,8 @@ public class PurpurConfig {
+         commands = new HashMap<>();
+         commands.put("purpur", new PurpurCommand("purpur"));
+ 
+-        version = getInt("config-version", 27);
+-        set("config-version", 27);
++        version = getInt("config-version", 28);
```
katekerllenevich: dont change this

katekerllenevich (changes requested):

patches/server/0272-Configurable-map-update-interval.patch
```diff
@@ -0,0 +1,44 @@
...
+--- a/src/main/java/org/purpurmc/purpur/PurpurConfig.java
++++ b/src/main/java/org/purpurmc/purpur/PurpurConfig.java
+@@ -427,6 +427,11 @@ public class PurpurConfig {
+         tpsCatchup = getBoolean("settings.tps-catchup", tpsCatchup);
+     }
+ 
++    public static int mapUpdateInterval = 5;
++    private static void mapUpdateInterval() {
++        mapUpdateInterval = getInt("settings.map-update-interval", mapUpdateInterval);
++    }
```
BillyGalbreath: this could be per-world if you use the player's world

BillyGalbreath (changes requested):

patches/server/0272-Configurable-map-update-interval.patch
minhh2792: Reverted

patches/server/0272-Configurable-map-update-interval.patch
minhh2792: I think it is better to leave it in global config, I don't think anyone would want long update interval anyway :p

BillyGalbreath: you'd be surprised. one thing i've learned with purpur over the years is no one wants anything until they realize its possible. its a simple edit to make it per-world compatible, so might as well do that before people get used to it being global and we have to make a migration for it ;)

minhh2792: yeah, i will work on it :)

minhh2792:
Moved to per world config, everything has been tested. This PR is now ready to merge

BillyGalbreath (approved):

katekerllenevich (approved):

kev626:
This patch makes no sense and should not be included.

First, you can see in this patch, if the map contains dirty data, then it will always be immediately rendered with no delay. Configuring the update interval will have no noticeable impact on performance nor will it have any noticeable impact on player appearance to the game due to this dirty check.

Second, there's no way your players are noticing quarter-second delay in maps, even if it were happening.

minhh2792:
This patch is not meant to be "performance improvement" anyway

A Minecraft CS:GO server that utilizes the Map, and 5 ticks delay per update is really slow and noticeable  

Edit link: https://youtu.be/i_0NrTb2Rsc

kev626:
> This patch is not meant to be "performance improvement" anyway
> 
> A Minecraft CS:GO server that utilizes the Map, and 5 ticks delay per update is really slow and noticeable
> 
> Edit link: https://youtu.be/i_0NrTb2Rsc

Then do you explain the presence of the “dirty” check?

minhh2792:
What dirty check

minhh2792:
This patch has nothing to do with the check tho, it just make the interval configurable

kev626:
> What dirty check

`if (!this.dirtyData && …)`
Did you even read the patch you were modifying?

minhh2792:
> This patch has nothing to do with the check tho, it just make the interval configurable

kev626:
> This patch has nothing to do with the check tho, it just make the interval configurable

This is why I think you didn’t even read the patch you’re attempting to modify. If the data used to render the map is changed, the map already updates immediately. In this sense, the interval doesn’t even matter.

The interval only serves to update the map data if for some reason the data is changed without the dirty flag being set (which should be never). Having the interval configurable at all is totally useless, as it’s value does not have any impact whatsoever on gameplay.

minhh2792:
I appreciate the clarification, in my case, the map (you can watch the video, it's faster) is slow. Changing the value works for me so I make this PR, maybe other servers also having this problem as well

If this PR is closed then I will just port it to my fork :fork:

amadeusmz:
@kev626 It will not be immediately rendered, it will check for dirty every 5 ticks as I told you on Pufferfish Discord.
This patch worked for me and I really need it.

kev626:
> @kev626 It will not be immediately rendered, it will check for dirty every 5 ticks as I told you on Pufferfish Discord.
> 
> This patch worked for me and I really need it.

If you need a map to be immediately rendered to a player, you can do this via the API by calling Player#sendMap(MapView). There is no need for any patch.

Regardless, it's not my project and purpur is free to add whatever they like, but this is truly not necessary.

cittyinthecloud:
This patch actually fixes a mismatch between API behavior in Spigot and Paper. Perhaps this would be better brought up upstream?

PR #950, opened 2022-03-31
Add suppress offline mode warning

Kale-Ko:


Kale-Ko:
Why?

BillyGalbreath:
We don't suppress useful information

Kale-Ko:
How is it useful?

Kale-Ko:
Its just an annoyance when you are running a bungee server

katekerllenevich:
Its so we know who not to provide support for.

BillyGalbreath:
Join us on Discord if you'd like to have a conversation about it (or search it up in there to see previous discussions on it).

Kale-Ko:
> Its so we know who not to provide support for.

Ah, fair enough. Thanks anyways

BillyGalbreath:
That's actually not any of the valid reasons we don't do this. We will only suppress pointless information, not useful information. And these offline messages (whether the server is in offline mode for piracy, testing, or behind a proxy) are very important information the server admins need to read and be aware of. It doesn't matter how annoying it is, it is important information, so it will never be suppressed by us.

PR #957, opened 2022-04-14
Config to prevent horses from standing when hurt

galacticwarrior9:
This PR adds a config option to toggle horses halting and standing on their hind legs when their hurt sound plays.

This behaviour can be undesirable to servers who wish to improve mounted combat. There is currently no way to prevent this behaviour through the API. 

Cactus and sweet berry bush damage will still cause the horse to rear. It's not clear to me why - input/changes are appreciated. Otherwise everything appears to work as intended.


patches/server/0272-Config-to-prevent-horses-from-standing-when-hurt.patch
```diff
@@ -0,0 +1,42 @@
...
+diff --git a/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java b/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java
+index f7088aaf65ee5725e928845a634852119b403cef..1106bc27b0d08109c93b7a6fab307f801f3dcd56 100644
+--- a/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java
++++ b/src/main/java/net/minecraft/world/entity/animal/horse/AbstractHorse.java
+@@ -386,7 +386,7 @@ public abstract class AbstractHorse extends Animal implements ContainerListener,
+     @Nullable
+     @Override
+     protected SoundEvent getHurtSound(DamageSource source) {
+-        if (this.random.nextInt(3) == 0) {
++        if (this.random.nextInt(3) == 0 && this.level.purpurConfig.horseStandWhenHurt) { // Purpur
```
encode42: Add the config check before the random check, we don't want to run unnecessary logic

encode42 (changes requested):

patches/server/0272-Config-to-prevent-horses-from-standing-when-hurt.patch
galacticwarrior9: Change made

encode42 (approved):
LGTM, just needs additional testing and a potential rebase

katekerllenevich (approved):
lgtm

PR #958, opened 2022-04-16
requirement in issues 

SageSphinx63920:
I added a requirement in the issues that offline mode servers aren't supported by purpur. Im not sure if this should be an option for itself or in () behind the supported option


granny:
This falls under `My version of Minecraft is supported by Purpur.`, and doesn't require clarification.

PR #961, opened 2022-04-24
Adding russian characters in nickname support

BrokenFG:
This PR allows players with russian characters in the nickname to connect to the server


katekerllenevich:
Does Mojang allow this in their usernames?

YouHaveTrouble:
> Does Mojang allow this in their usernames?

I'm 99.9% sure they do not. Only english alphanumeric characters and `_` as far as I know.

BrokenFG:
Most Russian servers support Russian nicknames, but there is no possibility to connect with such a nickname in this fork. I think there should be such an opportunity.

encode42:
As Russian usernames are only possible on offline mode servers and with cracked players, this will not be accepted. Purpur does not support offline mode, nor will we add features that apply specifically to it.

PR #966, opened 2022-04-30
Add toggle for enchant level clamping

12emin34:
By default, enchantment levels are stored as bytes and clamped to level 255, this patch allows storing them as shorts which would allow levels to go to 32767.

NOTE: the client will not display levels higher than 255.


katekerllenevich (approved):
lgtm

PR #974, opened 2022-05-06
Add support for Proxy Protocol

PanSzelescik:
This PR adds support for Proxy Protocol in Purpur.
More info in Paper PR, which is still open: https://github.com/PaperMC/Paper/pull/7710


katekerllenevich:
Possible problem, if paper merged this then we have to maintain a config migration for god knows how long on this.

PanSzelescik:
Right, is it better to add config to Paper? Not to Purpur?

katekerllenevich:
I'll be honest I'm not certain, would prefer advice from others

encode42:
IMO low-level changes such as this are more fit in Paper

katekerllenevich:
I agree

encode42:
As discussed, this feature is a better fit for Paper. If your upstream PR is not accepted, this PR can be revisited!

PR #1033, opened 2022-07-01
Add patch to prevent block desyncs

spookyoctopus:
It is quite common to have blocks get desynced between the client and
server, Especially with the use of plugins. 

This patch reduces the consequences and likely hood of such a desync, by sending a BlockUpdatePacket on a block
action such as digging.


YouHaveTrouble:
no com

PR #1034, opened 2022-07-01
Split allowUnsafeEnchants setting

Krakenied:
`allowInapplicableEnchants` - allows applying, for example, sharpness to a pickaxe
`allowIncompatibleEnchants` - allows applying, for example, protection and fire protection at the same time
`allowHigherEnchantsLevels` - allows exceeding enchants' maximum levels, for example: Eff V + Eff V = Eff VI


SageSphinx63920:
I think it's better to disable this option as default, like it's in the version right now.

Krakenied:
> I think it's better to disable this option as default, like it's in the version right now.

It's already disabled by default, but it doesn't work yet.

PR #1050, opened 2022-07-12
Chat Preview API

BillyGalbreath:
Opening this as a PR to get feedback before merging in.

Speak now, or forever hold your peace ;)

### How this works

1. The client will send a chat preview packet to the server with every keystroke into the chat textbox containing a query string of the textbox contents.
2. The server will try to decorate this query and return it to the client.
3. The client will determine if the decorated text differs from the text box contents when the chat message is sent (enter key pressed).
    * If the decoration differs, the signature is made from the contents of the decoration.
    * If the text is the same, the signature is made from the contents of the original message.
4. The server will receive the chat message and try to decorate the message again
    * If the decoration differs, the signature of the decoration is used to verify the message
    * If the text is the same, the signature of the original message is used to verify the message
5. The server then displays the original message in the chat box.
    * _Note that because of this, your preview should match whatever your chat event does for the desired results._

### Example usage

This is a simple plugin to convert raw input to minimessage, do not do this on production..

```java
public class TestPlugin extends JavaPlugin {
    @Override
    public void onEnable() {
        getServer().getPluginManager().registerEvents(new Listener() {
            @EventHandler
            public void on(PlayerPreviewChatEvent event) {
                event.setQuery(handleChat(event.getQuery()));
            }

            @EventHandler
            public void on(AsyncChatEvent event) {
                event.message(handleChat(event.message()));
            }

            private Component handleChat(Component component) {
                String raw = LegacyComponentSerializer.legacyAmpersand().serialize(component);
                return MiniMessage.miniMessage().deserialize(raw);
            }
        }, this);
    }
}
```

image

image


YouHaveTrouble:
Hopefully we can get signature verification working, sooner or later. I hope this PR comes through soonish, can't wait to mess with previews ;)

LOOHP:
Is that even possible considering that only the client can sign messages?

YouHaveTrouble:
> Is that even possible considering that only the client can sign messages?

Client gets the preview packet and signs it, then sends back the signed component message afaik

LOOHP:
> Client gets the preview packet and signs it, then sends back the signed component message afaik

However once the server tempers with that signed message it won't be accepted by other clients as signed.

YouHaveTrouble:
> However once the server tempers with that signed message it won't be accepted by other clients as signed.

but it doesn't. client gets how the message's supposed to look and signs that, then sends that resulting message back to the server which then sends it to other players

LOOHP:
> but it doesn't. client gets how the message's supposed to look and signs that, then sends that resulting message back to the server which then sends it to other players

But from the looks of this API, the preview and the actual chat message are handled in separate events. There is no guarantee that the preview sent to the client (which is signed) and the resulting chat message broadcasted are the same.

[Edit: I think what I'm thinking is not in the scope of this problem here]

BillyGalbreath:
I've already fixed the chat signing. I just haven't pushed it up yet.

granny (approved):

encode42 (approved):
Haven't tested, but code looks good to me

YouHaveTrouble (approved):
Tested for a little bit, seems to work just fine.

katekerllenevich (approved):

PR #1057, opened 2022-07-17
Remove Timings and Mojang's Profiler

BillyGalbreath:


encode42:
We should probably add some more information to our docs before this is merged.

Here's what I had in mind:
- [ ] A "tl;dr" of Spark, linking to Spark's docs for more info
  - Should explain the basics of how to use Spark's essential commands
  - Contains snippets of commands for various uses (timeout, ticks over 50, etc.)
  - Explain the major differences and why they do/don't matter (for ex. timings' time chart)
  - *"How to do X from Timings in Spark"* (find plugin lag, datapack lag, etc.)
- [ ] Why we're doing this
  - Performance implications and inaccuracies of Timings
  - Why Timings is never truly disabled unless removed entirely
  - Why Spark is a worthy successor
- [ ] Canned responses
  - Update `.spark` to explain what spark is, how to use it, and link to docs
  - Create `.send-spark` to hold what's currently `.spark`, or whatever snippets we need

katekerllenevich (approved):

patches/api/0048-Remove-Timings.patch
```diff
@@ -0,0 +1,151 @@
...
+-        this.usageMessage = "/timings <reset|report|on|off|verbon|verboff>";
++        this.usageMessage = "/timings";// <reset|report|on|off|verbon|verboff>"; // Purpur
+         this.setPermission("bukkit.command.timings");
+     }
+ 
+@@ -53,6 +53,13 @@ public class TimingsCommand extends BukkitCommand {
+         if (!testPermission(sender)) {
+             return true;
+         }
++        if (true) {
```
YouHaveTrouble: I'm curious about the origin story of that lonely `if (true)`

SageSphinx63920:
I think the old timings command should be replaced with a notice about the complete removal instead of being disabled.

katekerllenevich:
The timings command has been replaced with a notice.

BillyGalbreath:
> I think the old timings command should be replaced with a notice about the complete removal instead of being disabled.

That is what I've done here.

SageSphinx63920:
Oh sorry I only viewed the second commit of the pr, not the first

PR #1072, opened 2022-07-23
Add ability to place and remove block debug markers for players

YouHaveTrouble:
Debug markers exist so we can visualize the locations in minecraft world. Plugins could use it to debug issues with block locations, help with labeling blocks with normally hidden data (maybe blocks that have added interaction from a plugin, etc).

This *can* be also used to mark a block for player, but it's not recomended as the display is not consistant on all client versions, for example current 1.19 version ignores highlight color unless fixed with a datapack or mod. This fact is pointed out in javadoc.


patches/server/0292-debug-marker-api.patch
```diff
@@ -0,0 +1,60 @@
...
+     }
++
++    @Override
++    public void sendBlockHighlight(Location location, String text, int time, org.bukkit.Color color, int transparency) {
++        if (this.getHandle().connection == null) return;
++        io.netty.buffer.ByteBuf packet = io.netty.buffer.Unpooled.buffer();
++        //write location
++        packet.writeLong(((long) location.getX() & 67108863L) << 38 | (long) location.getY() & 4095L | ((long) location.getZ() & 67108863L) << 12);
++        //write color
++        packet.writeInt(new java.awt.Color(color.getRed(), color.getGreen(), color.getBlue(), transparency).getRGB());
```
BillyGalbreath:
```suggestion
+        packet.writeInt((transparency << 24) | (color.asRGB() & 0xFFFFFF));
```

BillyGalbreath: I would actually prefer to handle color as an int rather than a Bukkit or AWT Color so we're not forcing plugins to allocate an entire Object just for a color.

For example:
```
sendBlockHighlight(location, "text", time, 0x6400FF00); // lime with "100" transparency
```

YouHaveTrouble: I left one method that uses bukkit Color in constructor to give a little tease and motivation to learn color ints to use the rest of the methods ;P

patches/api/0050-debug-marker-api.patch
```diff
@@ -0,0 +1,62 @@
...
++    public void sendBlockHighlight(@NotNull Location location, int time, int color);
++
++    /**
++     * Creates debug block highlight on specified block location. Clients may be inconsistant in displaying it.
++     * @param location Location to highlight
++     * @param time Time after which highlight will expire in milliseconds
++     * @param color Color of the highlight. Will be ignored on some versions of vanilla client
++     * @param transparency Transparency of the highlight. Values not in range of 0-255 will throw IllegalArgumentException.
++     */
++    public void sendBlockHighlight(@NotNull Location location, int time, @NotNull org.bukkit.Color color, int transparency);
```
BillyGalbreath: No override with text and bukkit color? I'd recommend as many overrides as possible.

BillyGalbreath (changes requested):

BillyGalbreath (approved):

PR #1086, opened 2022-07-29
[ci-skip] Very important documentation fix

wzieciak:
Updated version of purpur api in docs.


YouHaveTrouble (approved):
I bet it will improve performance too!

PR #1133, opened 2022-09-26
Add an option to disable the TPS limit

MelnCat:
This adds an option to disable the TPS limit by disabling the waitUntilNextTick method. This allows for more than 20 ticks per second.


SageSphinx63920:
I don't think this is really useful for anyone lmao

MelnCat:
It's a fun, new, exciting gameplay feature

RhythmicSys:
Seems very essential to me, we need as much speed as possible

PR #1141, opened 2022-10-01
Implement squid colors for rainglow fabric mod

BillyGalbreath:


YouHaveTrouble (approved):

encode42 (approved):

triphora:
Purpur is now officially based

patches/server/0303-Implement-squid-colors-for-rainglow-fabric-mod.patch
```diff
@@ -0,0 +1,128 @@
...
++        ARO_PRIDE(BLACK, GRAY, WHITE, GREEN),
++        ENBY_PRIDE(YELLOW, WHITE, BLACK, PURPLE),
++        GENDERFLUID(PURPLE, WHITE, BLACK, PINK, BLUE),
++        MONOCHROME(BLACK, GRAY, WHITE),
++        VANILLA(BLUE);
++
++        private static final Map<String, Mode> BY_NAME = new HashMap<>();
++
++        static {
++            Arrays.stream(values()).forEach(mode -> BY_NAME.put(mode.name(), mode));
```
eristhea, line 105: @BillyGalbreath I believe that this is problematic
when I was implementing this optimisation myself the lookup failed every time, and I found out that that's because this line (exactly copied) used `.name()`, which will return the uppercase string ("VANILLA") instead of how it's stored in the config ("vanilla"). I don't know the details of how purpur's code works but I just thought I'd let you know.

eristhea: side note: I fixed this by using `toString()` instead of `name()`
https://github.com/ix0rai/rainglow/commit/1d2a437880190428123edd4f1c476b57bcea2144

BillyGalbreath: I noticed that after I published this. :3 However, it's not _really_ a problem, because a few lines below I am capitalizing the input in `get(String)`.

In hindsight, yeah, I should have used `toString()` and worked with lowercase inputs. But in the end, it doesn't really matter as long as it all matches.

I did test this thoroughly on my server with your mod, and it does work as expected ^_^

eristhea: perfect!
I might bother you again when I hit 1.0, as there will be a few new modes and colours and such to pop in

PR #1143, opened 2022-10-02
Configurable block fall damage modifiers

MelnCat:
This allows the block fall damage modifiers (distance and damage) to be configured.
For example, a configuration of 
```yaml
  block-fall-multipliers:
    minecraft:granite:
      distance: 10
    minecraft:oak_leaves:
      damage: 10
    minecraft:obsidian:
      damage: 100
      distance: 0.1
```
Causes falling (or jumping) on granite to act as if you've fallen 10x that distance, and increases the damage taken to be the damage taken if you've fallen 10x that distance.
Falling on oak leaves will increase damage by 10x, and obsidian will increase damage by 100x, but cause effective distance to be 10x less.


YouHaveTrouble (approved):

patches/server/0306-Configurable-block-fall-damage-modifiers.patch
```diff
@@ -0,0 +1,113 @@
...
+diff --git a/src/main/java/org/purpurmc/purpur/PurpurConfig.java b/src/main/java/org/purpurmc/purpur/PurpurConfig.java
+index 14844cb8a402ac987f44a91b95d5ebdac73c6dfe..b59ea8df8171f2d5b388d2f6f5619513f31cd1f1 100644
+--- a/src/main/java/org/purpurmc/purpur/PurpurConfig.java
++++ b/src/main/java/org/purpurmc/purpur/PurpurConfig.java
+@@ -588,4 +588,53 @@ public class PurpurConfig {
+             block.explosionResistance = blastResistance.floatValue();
+         });
+     }
++    private static void blockFallMultiplierSettings() {
++        Map<String, Float> defaultHayBlockMultiplier = Map.of(
```
frog-commits: reason for defining unlike the rest?

MelnCat: Originally defined beds too but the yaml serializer didn't like it, I'll go inline the hay block too

MelnCat: Actually, it doesn't really matter

frog-commits (approved):
all good, was just curious ☻

katekerllenevich (approved):

encode42 (approved):

PR #1149, opened 2022-10-10
Advancements rewards exp toggle

RealRONiN:
This feature can be used by RP servers who use experience as currency etc.


granny:
Closing this PR since you can modify the rewards an advancement gives by creating a datapack.

PR #1153, opened 2022-10-11
Ajustable burp delay

5jd:
Basically a ajustable burp delay config option

Edit: Renamed Cooldown to Delay after Billy's recommandations


katekerllenevich (approved):

BillyGalbreath (changes requested):
This field was added by Purpur and was never really intended to be configurable. All it does it delay the burp sound from playing until the final "eat" sound is done (or almost done) playing on the client so the burp can actually be heard.

If we're going to make this configurable, I think we should rename the field to something better suited for it's intended purpose. Like, maybe `burpDelay`.

This should also be merged into the original patch.

5jd:
> This field was added by Purpur and was never really intended to be configurable. All it does it delay the burp sound from playing until the final "eat" sound is done (or almost done) playing on the client so the burp can actually be heard.
> 
> If we're going to make this configurable, I think we should rename the field to something better suited for it's intended purpose. Like, maybe `burpDelay`.
> 
> This should also be merged into the original patch.

Got it!
I'm gonna make the changes whenever I can. :)

BillyGalbreath (approved):

PR #1155, opened 2022-10-11
Isn't the performance also built on top of pufferfish?

Soapy7261:
I mean, i could be wrong but since purpur has patches from pufferfish i thought it would also be built on top pufferfish (at least somewhat)


emielderckx:
Sure, we use the patches from Pufferfish. But Paper does the most work (starlight, dfu rewrite and chunk rewrite + many other optimalisations). Also more people know Paper so it just makes sense to have them there.

Soapy7261:
perhaps it should instead be paper & pufferfish?

katekerllenevich:
We aren’t built on top of Pufferfish like we are on Paper, it’s all in the semantics.

It’s the same thing how Pufferfish includes Petal patches yet claims to be built on top of Papaer.

PR #1160, opened 2022-10-14
Add log suppression for library loader

Krakenied:


BillyGalbreath:
Seems a bit excessive to expose it to API, is that needed?

BillyGalbreath (approved):

PR #1168, opened 2022-10-22
Add option to fix creative mode actions with NetworkItemSerializeEvent

MelnCat:
Currently in creative mode, items modified by NetworkItemSerializeEvent will persist even though they're client-sided due to the creative client using creative inventory actions. This pull request fixes that by adding a configuration options to send an additional item tag that stores the original unmodified item (that doesn't exist in the server item). When the creative player sends a creative inventory action to create the item with the tag, the server will convert it to the item stored in the tag.


granny (changes requested):
rebase & documentation required

granny (approved):
docs can wait

PR #1207, opened 2022-11-30
Fix the timeout for alternative keepalive handling

MartijnMuijsers:
The timeout for alternative keepalive handling uses KEEPALIVE_LIMIT, but that now uses milliseconds, so a factor 1000 is needed to make it have effect. In addition, >= is the equivalent to vanilla behavior rather than >, which will timeout after at least 30 seconds (instead of after at least 31).

(Also while we're making a change anyway we can replace List<Long> with LongList~)


BillyGalbreath (approved):
Huh. Looks like it's always been milliseconds. Guess that was a huge oversight on my part all these years :3

MartijnMuijsers:
Just a very tiny one then :p

PR #1208, opened 2022-11-30
Move PlayerCommandSendEvent listeners check

MartijnMuijsers:
So I apologize in advance for the tininess of this pull request that probably will never matter, but I think the PlayerCommandSendEvent listener check should be placed two lines lower because the AsyncPlayerSendCommandsEvent call is unrelated.


BillyGalbreath (approved):

BillyGalbreath:
No contribution to make things better is too small ^_^

PR #1240, opened 2023-01-02
[ci-skip] Change API version tp 1.19.3

SageSphinx63920:
We have an API for 1.19.3


encode42:
The same can be done for the docs

SageSphinx63920:
I opened a PR on the docs repo

PR #1250, opened 2023-01-15
Add option and permission to disable mining slowdown while flying

RhythmicSys:
This PR adds a purpur global config option: `flight.keep-mining-slowdown` - which defaults to true. When set to false, if a player is given the permission `purpur.flight.mining-slowdown-bypass`, they will not have the vanilla mining slowdown applied to them while they are in the air.

Vanilla slowdown multiplies the mining time by 5 when you are in the air

Current issues with this are that when the slowdown is removed, break particles and sounds do not play for the client. Subtitles do play, though. I am not sure how to fix this


granny:
Discussed in discord. Current patch only removes the slowdown when the item would be insta-mined without flight slowdown applied. Not the desired effect so closing this PR for now.

PR #1286, opened 2023-03-14
Fix NPE when no world given to explosions

BillyGalbreath:
Fixes compatibility issues when plugins somehow create explosions without a world object like BONNePlayground/DragonFights#18


encode42 (approved):

Olthoo:
Thanks for this fix !

PR #1290, opened 2023-03-19
Improve JS exec return value handling

skbeh:


katekerllenevich:
Thank you for the PR! You've come at just the worst timing as we just switched our master branch to `ver/1.19.4` so if you wouldn't mind rebasing the changes there that would be wonderful.

skbeh:
@theomega24 Rebased.

katekerllenevich:
Very nice, thanks for the PR!

PR #1292, opened 2023-03-20
Add option to disable scissor damage in liquid

JustDoom:
Added an option to disable scissor damage while in liquid. Some people found it annoying that they take damage when swimming clearing kelp. Made it optional just in case people like to get stabbed while swimming. It is enabled by default, think most players would not like it to be enabled

This is my first commit, please let me know if I did anything wrong


JanTuck:
It seems you included some extra patches with your suggested patch.

JustDoom:
ah just noticed. might have to fix that tomorrow as it's getting late for me

PR #1294, opened 2023-03-20
Fix the Optional type server status version being compared to null

MartijnMuijsers:
The server status version (`MinecraftServer.getServer().getStatus().version()`) used to be a nullable class field, but it is now an Optional record field. Therefore, the comparison to null should be replaced with `isEmpty()`. In its current state, with the null check, this patch does nothing anymore (because the version will never be null).

To verify the status will in fact be an empty Optional, see `net.minecraft.network.protocol.status#CODEC`, which uses `optionalFieldOf`, which returns a `com.mojang.serialization.codecs.OptionalFieldCodec` for which the `decode` method never returns null, but an empty Optional in the corresponding scenarios instead.


granny (approved):
Seems like it was an oversight during the update to 1.19.4 :) Thanks!

PR #1295, opened 2023-03-21
Add option to disable scissor damage in liquid

JustDoom:
Added an option to disable scissor damage while in liquid. Some people found it annoying that they take damage when swimming clearing kelp. Made it optional just in case people like to get stabbed while swimming. It is enabled by default, think most players would not like it to be enabled

This is my first PR, please let me know if I did anything wrong. Sorry about the previous PR too, was trying to fix something and accidentally deleted it which closed the PR

### DIFF
--- a/patches/server/0134-Dont-run-with-scissors.patch
+++ b/patches/server/0134-Dont-run-with-scissors.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] Dont run with scissors!
 
 
 diff --git a/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java b/src/main/java/net/minecraft/server/network/ServerGamePacketListenerImpl.java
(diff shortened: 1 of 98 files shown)


BrycensRanch:
things that modify behavior are typically disabled by default

encode42:
That is true! The current defaults would result in an unexpected behavior change; scissors no longer doing damage in water. Easy fix though!

BrycensRanch:
Yeah! thank you for contributing to Purpur to make it more awesome 🌋

JustDoom:
Fixed!

granny (changes requested):
Looks good to me! It looks like you're checking both water & lava but the option is labeled as "water". My request is to create a separate option specifically for lava as well.

JustDoom:
Fixed it! There are now two options

granny (approved):
LGTM

PR #1297, opened 2023-03-22
Add options to disable tripwire hook and leave decay updates

SageSphinx63920:
This pr adds config option to disable both things.

### DIFF
--- a/patches/server/0083-Add-option-to-disable-certain-block-updates.patch
+++ b/patches/server/0083-Add-option-to-disable-certain-block-updates.patch
@@ -145,30 +145,22 @@ index 50ead76879398222a76f26c36e98800d2d1af167..42e8e2d9d673947ecbfb55e0f31fc22a
          }
  
 diff --git a/src/main/java/net/minecraft/world/level/block/TripWireBlock.java b/src/main/java/net/minecraft/world/level/block/TripWireBlock.java
-index 7f60175bf671d282c11e9084670d2bb900968255..aed4cb2eff5f48279333dec00246d211fee8a00a 100644
+index 7f60175bf671d282c11e9084670d2bb900968255..ea0a3fb7fb325707a01aa619ace4c33fcb522131 100644
 --- a/src/main/java/net/minecraft/world/level/block/TripWireBlock.java
 +++ b/src/main/java/net/minecraft/world/level/block/TripWireBlock.java
-@@ -24,6 +24,7 @@ import net.minecraft.world.level.gameevent.GameEvent;
- import net.minecraft.world.phys.shapes.CollisionContext;
- import net.minecraft.world.phys.shapes.VoxelShape;
- import org.bukkit.event.entity.EntityInteractEvent; // CraftBukkit
-+import org.purpurmc.purpur.PurpurConfig;
- 
- public class TripWireBlock extends Block {
- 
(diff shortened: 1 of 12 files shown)


SageSphinx63920:
I pr on the docs when this is approved

SageSphinx63920:
The action fails, because the pufferfish/some random patch before it patch did smth bad. Running `./gradlew applyPatches` works fine and `./gradlew runDev` works also without any errors

patches/server/0083-Add-option-to-disable-certain-block-updates.patch
```diff
@@ -98,6 +98,26 @@ index 3c6d97b51c6fec130b80e5965afa2c49d48843c9..b456cb8efd8f0be8a6860c82462ce9bd
...
+             LeavesDecayEvent event = new LeavesDecayEvent(world.getWorld().getBlockAt(pos.getX(), pos.getY(), pos.getZ()));
+             world.getCraftServer().getPluginManager().callEvent(event);
+@@ -65,6 +66,7 @@ public class LeavesBlock extends Block implements SimpleWaterloggedBlock {
+     }
+ 
+     protected boolean decaying(BlockState state) {
++        if(org.purpurmc.purpur.PurpurConfig.disableLeavesDecay) return false; // Purpur - disable leaves decay
+         return !(Boolean) state.getValue(LeavesBlock.PERSISTENT) && (Integer) state.getValue(LeavesBlock.DISTANCE) == 7;
+     }
+
```
granny, line 120: unneeded since you can set `persistent` to the blockstate to stop it from decaying

patches/server/0083-Add-option-to-disable-certain-block-updates.patch
```diff
@@ -124,21 +144,45 @@ index 50ead76879398222a76f26c36e98800d2d1af167..42e8e2d9d673947ecbfb55e0f31fc22a
...
+         if (!world.isClientSide) {
+             if (!(Boolean) state.getValue(TripWireBlock.POWERED)) {
+@@ -143,6 +144,7 @@ public class TripWireBlock extends Block {
+     }
+ 
+     private void checkPressed(Level world, BlockPos pos) {
++        if(org.purpurmc.purpur.PurpurConfig.disableTripwireUpdates) return; // Purpur - disable tripwire updates
+         BlockState iblockdata = world.getBlockState(pos);
+         boolean flag = (Boolean) iblockdata.getValue(TripWireBlock.POWERED);
+         boolean flag1 = false;
```
granny, line 166: checks not placed in the correct methods

granny (changes requested):
Remove the `disable-leaves-decay-updates` option completely, it's unneeded.

Your changes to `TripWireBlock` don't do what you think it does. Remove what you have and take inspiration from the other changes in this patch.

Hint: all public overridden methods that are meant to return a `BlockState` include a check that returns a default blockstate of the block.

granny:
> The action fails, because the pufferfish/some random patch before it patch did smth bad. Running `./gradlew applyPatches` works fine and `./gradlew runDev` works also without any errors

should be fine once you merge changes from `ver/1.19.4`

granny:
any updates? this pull request will be closed if it continues to be stale

SageSphinx63920:
Yeah sorry granny, I need to be reminded that sth like this exits bc I forgot about it. Ill fix it tomorrow

PR #1299, opened 2023-03-25
Option to damage players by throwing an anvil item

12emin34:
This patch allows damaging players by throwing anvil items at them, this doesn't work for other entities tho, as I haven't really figured out a way to do it.


RealRONiN:
This should also work with Dripstone, could you check and maybe add that aswell? do rename the PR if you're going to add that

GhostIsBeHere:
This might be better to put in PurpurExtras imo

Issue #1344, opened 2023-06-12
ContainerAnvil m() throws NullPointerException


FloGamerMC:
### Spark link

don't have it

### Expected behavior

I don't know what this method do, because it's a [library](https://github.com/WesJD/AnvilGUI/blob/master/1_20_R1/src/main/java/net.wesjd.anvilgui.version/Wrapper1_20_R1.java#L96). But I tested it with the newest Paper version and it worked. Maybe you must add the newst paper patches. 

### Observed/Actual behavior

The method throws an exception:
```
java.lang.NullPointerException: Cannot invoke "String.length()" because "input" is null
	at net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializerImpl.deserialize(LegacyComponentSerializerImpl.java:207) ~[adventure-text-serializer-legacy-4.14.0.jar:?]
	at net.minecraft.world.inventory.AnvilMenu.createResult(AnvilMenu.java:342) ~[?:?]
	at net.wesjd.anvilgui.version.Wrapper1_20_R1$AnvilContainer.m(Wrapper1_20_R1.java:96) ~[Fuxelhunt.jar:?]
	at net.minecraft.world.inventory.ItemCombinerMenu.slotsChanged(ItemCombinerMenu.java:112) ~[?:?]
	at net.minecraft.world.inventory.ItemCombinerMenu$3.setChanged(ItemCombinerMenu.java:103) ~[?:?]
	at net.minecraft.world.SimpleContainer.setItem(SimpleContainer.java:207) ~[?:?]
	at org.bukkit.craftbukkit.v1_20_R1.inventory.CraftResultInventory.setItem(CraftResultInventory.java:37) ~[purpur-1.20.jar:git-Purpur-1990]
	at net.wesjd.anvilgui.AnvilGUI.openInventory(AnvilGUI.java:170) ~[Fuxelhunt.jar:?]
	at net.wesjd.anvilgui.AnvilGUI.access$1000(AnvilGUI.java:37) ~[Fuxelhunt.jar:?]
	at net.wesjd.anvilgui.AnvilGUI$Builder.open(AnvilGUI.java:561) ~[Fuxelhunt.jar:?]

```

### Steps/models to reproduce

I used this library AnvilGUI, on the Point I want to open the anvil gui this exception came up. But as mentioned above with the newest paper build it works fine.

### Purpur version

[19:12:34 INFO]: Checking version, please wait...
[...]

PR #1350, opened 2023-06-14
AnvilMenu createResult() set itemName to "" if name is null

FloGamerMC:
The LegacyComponentSerializer threw an NullPointerException because the name was null(This could be expected because the itemName have an @Nullable annotation).  Now it is an empty string.
The PR is a fix for my issue https://github.com/PurpurMC/Purpur/issues/1344 .


granny:
The underlying issue was most likely a mistake on our part when we updated purpur to 1.20. Appreciate the PR nonetheless!

PR #1361, opened 2023-06-23
Delete cache on build failure

TheRealRyGuy:
cc granny
This deletes cache on build fail, still requires a rerun though 
If you wanted to combine the rerun, just needs a check (label?) to ensure it only runs once
`POST /repos/:owner/:repo/actions/runs/:run_id/rerun` 
https://docs.github.com/en/rest/actions/cache?apiVersion=2022-11-28 - cache API

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,6 @@
 name: Build
 on:
+  workflow_dispatch:
   pull_request:
   push:
     branches:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,3 +23,8 @@ jobs:
         run: ./gradlew applyPatches --no-daemon --stacktrace
       - name: Build
         run: ./gradlew build --no-daemon --stacktrace
+      - name: Delete cache on Failure
+        if: ${{ failure() }}
+        uses: mokmok-dev/delete-actions-cache@main
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


BillyGalbreath:
why not just delete the cache before every run? :sipjuice:

image

granny:
because i'm extra

TheRealRyGuy:
i lowkey don't like the impl of that cache action so i may make my own 
but this should do part of the trick - auto rerunning feels sketchy cause we can't check the error message

TheRealRyGuy:
k this will have actual debug now when clearing cache and is readable & easier to modify 
edit: still doesn't build that yaml library smh

granny (commented):
Proposal: Instead of clearing the actions cache, what if we just copy/pasted the "Apply Patches" & "Build Patches" steps but with the failure condition on both, as well as a step before that runs something like `./gradlew clean cleanCache`.

So essentially:

1. Apply Patches
2. Build Patches
3. If failed, run `./gradlew clean cleanCache`
4. If failed, run "Apply Patches"
5. If failed, run "Build Patches"

TheRealRyGuy:
i mean clean (should?) depend on cleanCache, i'm not against fully clearing the actions cache either way as well 
not a bad idea though to run it like that 
- side note the cache fix doesn't fix building, that's #1364

TheRealRyGuy:
forgot abt this my fault - fixed it though. no longer deletes actions cache

granny (approved):

PR #1372, opened 2023-07-02
Shears can defuse TNT

Findus642:
TNT can now be defused when right clicking with shears, nothing more to it than an additional gameplay mechanic. 

Option is per-world and a value of 0.00 disables the function. See patch commit for a few more details

### DIFF
--- a/patches/server/0306-Shears-can-defuse-TNT.patch
+++ b/patches/server/0306-Shears-can-defuse-TNT.patch
@@ -44,26 +44,10 @@ index bf3301eb1341ba9d482e10873447c42bd670f5ed..3c46956efec714a2286ad8bc058f43a3
 +    // Purpur end - Shears can defuse TNT
  }
 diff --git a/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java b/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java
-index 64527770986fb44cb2b569924b2f98c9570b25b2..ec39dc4bbbde346ba404c46c15c75d2d8fa88db3 100644
+index 64527770986fb44cb2b569924b2f98c9570b25b2..59c413b5d6681da2636d7cfa0d452a4643e9f84b 100644
 --- a/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java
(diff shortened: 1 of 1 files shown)


patches/server/0306-Shears-can-defuse-TNT.patch
```diff
@@ -0,0 +1,77 @@
...
+ }
+diff --git a/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java b/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java
+index 64527770986fb44cb2b569924b2f98c9570b25b2..ec39dc4bbbde346ba404c46c15c75d2d8fa88db3 100644
+--- a/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java
++++ b/src/main/java/org/purpurmc/purpur/PurpurWorldConfig.java
+@@ -8,7 +8,6 @@ import net.minecraft.world.entity.EntityType;
+ import net.minecraft.world.item.DyeColor;
+ import net.minecraft.world.item.Item;
+ import net.minecraft.world.item.Items;
+-import net.minecraft.world.level.Explosion;
```
granny: needs to be reverted

patches/server/0306-Shears-can-defuse-TNT.patch
```diff
@@ -0,0 +1,77 @@
...
+ import net.minecraft.world.level.block.Block;
+ import net.minecraft.world.level.block.Blocks;
+ import net.minecraft.world.level.block.state.properties.Tilt;
+@@ -23,7 +22,6 @@ import org.bukkit.configuration.ConfigurationSection;
+ import java.util.ArrayList;
+ import java.util.HashMap;
+ import java.util.List;
+-import java.util.Locale;
```
granny: needs to be reverted.

granny (changes requested):
Currently on mobile so i can't look into the implementation, so nitpicking for now until I can take a better look. Patch diffs should touch as little as possible to lower the maintenance burden of each patch.

Findus642:
Both been reverted. Didn't pick up on this during committing!

granny (approved):
LGTM :)

All that's left is to document the config option and permission in the docs (https://github.com/purpurMC/purpurdocs) and the patch will be merged!

PR #1418, opened 2023-08-23
Add option to always show item in player death messages

MelnCat:
Currently, the item in player death messages is only shown if it is renamed, i.e. Player was killed by Player2 using [cool sword].
This patch adds a config option to allow always having the item show up, so if a player is killed with a non-renamed diamond sword, the death message will note that the killing weapon was a diamond sword.


granny (changes requested):
- the config option should be checked first before the `ItemStack#hasCustomHoverName()` method
- please PR to PurpurDocs the new config option

granny (approved):

PR #1419, opened 2023-08-23
Add configuration option to replace incompatible enchantments

MelnCat:
Currently, the behavior of incompatible enchantments in anvils is either to discard any enchantments that are incompatible, keeping the original enchantments (vanilla), or to allow both enchantments to exist (allow incompatible).
This pull request adds a third option: to replace the existing incompatible enchantments and to replace it with the applied enchantment at `settings.enchantment.anvil.replace-incompatible-enchants`

For example, an item with Sharpness and Unbreaking + an item with Smite and Fire Aspect will result in an item with Sharpness Unbreaking and Fire Aspect with vanilla behavior. With this configuration option, the output will instead be Smite + Unbreaking + Fire Aspect.


granny (approved):
LGTM! Just needs to be merged into `patches/server/0153-Config-to-allow-for-unsafe-enchants.patch` and documented in PurpurDocs

MelnCat:
Merged into the unsafe enchants patch

PR #1443, opened 2023-10-09
Add option to disable the copper oxidation proximity penalty

MelnCat:
Currently, copper will have a lower chance of being oxidized if there are copper blocks near it in a 4 block radius that are less oxidized than it. This required copper blocks to be spaced out to oxidize faster.
This change adds an option in the purpur world settings named `disable-oxidation-proximity-penalty` which when enabled, disables the proximity penalty so the chance of oxidation will be the same regardless of the presence of other copper blocks.


granny (approved):
LGTM! Just needs a doc PR :)

PR #1451, opened 2023-10-17
config for turning bundles into functional quivers

granny:
adds two new options; `gameplay-mechanics.item.bow.use-bundle-as-quiver` and `gameplay-mechanics.item.crossbow.use-bundle-as-quiver`. With either of these options enabled, you can use arrows that are stored inside a bundle like you would a quiver.

New Config options:
```yaml
gameplay-mechanics:
  item:
    bow:
      use-bundle-as-quiver:
        item: false
        component: false
    crossbow:
      use-bundle-as-quiver:
        item: false
        component: false
```

TODO:

- [ ] fix client-side visual bug where the client will think that an arrow in the player's inventory is being used instead of the one in the bundle.
- [ ] fix client-side visual bug when using the bundle with arrows in the offhand.
- [x] fix creative mode bug when the player continues to consume arrows in bundles when they shouldn't


DereC4:
a really fun idea!

PR #1489, opened 2024-02-18
Fix bee count payload lag/crash exploit and make it configurable.

dexuby:
Just a quick fix. This is generally a questionable feature, nothing about it is documented and it validates nearly nothing. Players can spam custom payload packets targeting locations anywhere in the world causing chunk loads resulting in freezing.


granny (changes requested):
1. Merge your patch into the `###-give-bee-counts-in-beehives-to-Purpur-clients.patch` patch
2. `PurpurConfig`, not `PufferfishConfig`
3. Needs to be documented at https://github.com/PurpurMC/PurpurDocs
4. Set it as `false` by default. [Someone requested this option a while ago](https://github.com/PurpurMC/Purpur/discussions/1434) and I don't mind it being disabled by default. Will need to double check that the PurpurClient mod doesn't freak out if it's disabled.

YouHaveTrouble:
If we're fixing things, let's fix them correctly. This shouldn't work unless player is in the range of the debug targeting range. This would prevent this becoming NoCom 2.0

dexuby:
A range check would be nice but I'm generally not too sure why this feature is necessary and what it's being used for (found nothing in the docs) so I didn't implement one. So this is being used by a Purpur mod to render some extra info?

YouHaveTrouble:
Yep. it's for PurpurClient feature that displays the bee count in F3 menu

dexuby:
Thanks, implemented a range limit of 16 now, that seems to be the max range when targeting blocks.

granny (approved):
Will merge later this afternoon (PST). Thanks for the PR!

granny:
GitHub CLI fucked up when I tried pushing to your PR so I've gone ahead and just merged the branch that was unintentionally created. Your changes have been merged.

PR #1509, opened 2024-04-28
Fix compilation issues

Krakenied:
- removed placeables/destroyables setters and getters as there's currently no sensible way of implementing it


Krakenied:
Somebody more into Purpur needs to take a look at the second half of the issues as I really don't know how to make some features still work after conversion from 1.20.4 to 1.20.5 (due to tags/components)

granny (approved):
Thank you for the help :)

PR #1517, opened 2024-05-26
Fix one tiny inconsistency in README

CyberFlameGO:
Gradle and Maven have inconsistent dependency numbers :sob:


granny (approved):
teehee oopsies

PR #1525, opened 2024-06-08
Switch to NogyangSpigot Fork

Outfit8TSB:
i checked out NogyangSpigot and the api patched correctly as a submodule.
from future purpur builds instead of Using Paper as a upstream, Use NogyangSpigot as a submodule.

Tasks:

- [ ] Fix Server Patches

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "NogyangSpigot"]
+	path = NogyangSpigot
+	url = https://github.com/GAME-CLI-SRV-DEV/NogyangSpigot
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -143,4 +143,5 @@ tasks.register("printPurpurVersion") {
     doLast {
         println(project.version)
     }
+ }
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -90,7 +90,7 @@ patchTasks {
                     outputDir = layout.projectDirectory.dir("Purpur-api")
                 }
                 register("server") {
-                    upstreamDir = paperDir.dir("NogyangSpigot-Server")
+                    upstreamDir = paperDir.dir("NogyangSpigot-server")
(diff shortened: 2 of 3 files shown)


Euphillya:
I don't think your PR will be accepted as it is, because we will be depending on another project that might be abandoned in the future. Otherwise, create a patch that implements their functionality as we did with:
- Tuinity
- Airplane
- PufferFish

YouHaveTrouble:
Not happening.

Outfit8TSB:
NOTE:  i don't think the patch is a good idea because NogyangSpigot Project will literally fall like Yatopia if we do that. (This is why i set it on draft.)

YouHaveTrouble:
Please refrain from opening PRs you don't want merged

Outfit8TSB:
Dude i was literally talking to euphillya. didn't expect that. If NogyangSpigot Implements Tuinity API, Airplane API, Pufferfish API. Pufferfish Server, Airplane Server but if we do that, we are going to collapse like "Yatopia". But i considered implementing tuinity + airplane + pufferfish in my project.

granny:
That's flawed from the start. Tuinity was merged into Paper, which means it's dead. Pufferfish is a continuation of Airplane, which means Airplane is obsolete. That leaves just Pufferfish (which still hasn't updated to 1.20.6)

Outfit8TSB:
Yes, in fact, if i do what euphilla said, We are going to more look like yatopia. See 
https://github.com/kennytv/yaptapia for more

Euphillya:
> Yes, in fact, if i do what euphilla said, We are going to more look like yatopia. See https://github.com/kennytv/yaptapia for more

I never said to make a Yatopia 2, just do it as it's been done for years like this: https://github.com/PurpurMC/Purpur/blob/ver/1.20.6/patches/unapplied-server/0001-Pufferfish-Server-Changes.patch

granny:
I see absolutely zero benefit in switching to your fork. Please refrain from bringing this up in the future.

PR #1530, opened 2024-06-15
Fix compilation issues

Krakenied:
it works now


Krakenied:
It would compile however there are references to bad timings methods so it just won't. (I leave it to be merged as I don't have more time to do that today)

granny (approved):
LGTM! Thank you for your help!

PR #1539, opened 2024-06-23
Improve AFK system

Dreeam-qwq:
* Add /afk command
* Add /afk command cooldown to prevent using /afk to spawm message, default 0 means no cooldown. (unit: second)
* Add configurable afk away title, can customize title and subtitle when afk away, and also customized time of title fade in, stay and fade out. (unit: ticks)


YouHaveTrouble:
This could be easily implemented via the available API without the additional maintenance burden

PR #1564, opened 2024-07-23
Add config to implement old raid behavior

2stinkysocks:
Added a config option `old-raid-behavior` to allow pre-1.21 raid farms to work as before, defaulting to false. Tested on ianxofour's stacking raid farm and ray's works' normal raid farm. The code was adapted from a fabric mod: https://github.com/litetex-oss/mcm-raid-restore/


2stinkysocks:
https://github.com/PurpurMC/PurpurDocs/pull/102

granny:
The original patch ended up making a ton of unnecessary changes that weren't needed to get the old raid behavior back. After doing some testing, the change I just pushed to this PR is the bare minimum needed to get this working.

With that being said, I believe this would be better implemented as a plugin instead. This can be done by listening to the `EntityDeathEvent` event to remove the ominous bottle from the entity drops while also applying the corresponding bad omen effect to the player. You'd also listen to the `EntityPotionEffectEvent` event for the `PotionEffectType` of `RAID_OMEN` to modify it to one second.

PR #1589, opened 2024-09-15
Fix break elytra

### LINKED ISSUE #1574: Elytra settings break elytras similarly to armor
### Spark link

https://spark.lucko.me/UYx0jOshLh

### Expected behavior

Elytras shouldn't break/disappear, they should only become disabled and unusable until repaired, but still stay in the player's chest plate slot/inventory. Not present on Paper, this is a bug that only happens on Purpur.

### Observed/Actual behavior

With Purpur 1.21 builds, I'm starting to notice that the following elytra settings result in cases where breaking the elytra instead of simply disabling them like in Vanilla occurs:
- Setting `damage-per-second:` to more than 1.0
- Setting `damage-multiplied-by-speed` to more than 1.0
- Setting any `damage-per-boost` setting to more than 0.0

### Steps/models to reproduce

By doing the following:
- Setting `damage-per-second:` to more than 1.0
- Setting `damage-multiplied-by-speed` to more than 1.0
- Setting any `damage-per-boost` setting to more than 0.0

And then getting elytras with near 0 durability, you can fly normally or with fireworks. Try variations of flying styles, like flying straight up and then gliding down at a 90 degree angle for maximum speed.

### Purpur version

purpur version
[15:37:10 INFO]: Checking version, please wait...
[15:37:11 INFO]: Current Purpur Version: 1.21-2280-930003b (MC: 1.21)*
Previous: 1.21-124-df3b654 (MC: 1.21)

### Agreements

- [X] I am running the latest version of Purpur available from https://purpurmc.org/downloads.

### Other

_No response_

Pantera07:
Closes #1574

This change is still needed, but has been removed: https://github.com/PurpurMC/Purpur/blob/0d6766e21d08f8348c666a460cb145fdc3a45ed0/patches/server/0057-Implement-elytra-settings.patch#L65-L70

### DIFF
--- a/patches/server/0060-Implement-elytra-settings.patch
+++ b/patches/server/0060-Implement-elytra-settings.patch
@@ -45,6 +45,23 @@ index 218f2f085309f04438f8b07bc41cf242583db2dc..ea8e49b42b9dde74784189430be66ed6
                      if (event.shouldConsume() && !user.hasInfiniteMaterials()) {
                          itemStack.shrink(1);
                      } else ((net.minecraft.server.level.ServerPlayer) user).getBukkitEntity().updateInventory();
+diff --git a/src/main/java/net/minecraft/world/item/ItemStack.java b/src/main/java/net/minecraft/world/item/ItemStack.java
+index e90cf9da9c35c5b73b489990141311c283db4767..59b54a0c37a435976f92a57f5838e25c87096213 100644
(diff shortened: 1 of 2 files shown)


granny (approved):

granny:
Thanks for the PR :)

PR #1594, opened 2024-09-30
Fix shulker box contents drop when destroyed

Dreeam-qwq:
Before the change, only purple shulker box can drop contents when destroyed (burned), and this pull request fix it.


granny:
I've just pushed a fix. Thank you for the PR nonetheless :)

PR #1621, opened 2024-12-02
add VaultStateChangeEvent

Grabsky:
Simplified version of https://github.com/PaperMC/Paper/discussions/11679. May extend this API further in the future but it is already good enough for use-case described in the discussion thread.

I don't know if I put the implementation in the right place because currently - I'm afraid - the event is also called when `Vault.State` is changed via API and doing so inside the event can create an indefinite loop if handled incorrectly. I'm not sure what should I do about this.

### DIFF
--- a/patches/api/0049-add-VaultStateChangeEvent.patch
+++ b/patches/api/0049-add-VaultStateChangeEvent.patch
@@ -1,0 +1,76 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Micha=C5=82=20Czopek?=
+ <44530932+Grabsky@users.noreply.github.com>
+Date: Mon, 2 Dec 2024 18:59:00 +0100
+Subject: [PATCH] add VaultStateChangeEvent
+
+
+diff --git a/src/main/java/org/purpurmc/purpur/event/block/VaultStateChangeEvent.java b/src/main/java/org/purpurmc/purpur/event/block/VaultStateChangeEvent.java
+new file mode 100644
+index 0000000000000000000000000000000000000000..d73a449cb4da3dec441242ae6db54a74a6f09a14
+--- /dev/null
++++ b/src/main/java/org/purpurmc/purpur/event/block/VaultStateChangeEvent.java
+@@ -0,0 +1,63 @@
++package org.purpurmc.purpur.event.block;
++
++import org.bukkit.block.Block;
++import org.bukkit.block.data.type.Vault;
++import org.bukkit.event.Cancellable;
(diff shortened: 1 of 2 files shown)


Onako2:
Is there a reason behind not creating a PR at Paper?
Just asking because Paper is more wide-spread and would profit more people

Grabsky:
> @Onako2: Is there a reason behind not creating a PR at Paper?
> Just asking because Paper is more wide-spread and would profit more people

Linked discussion have not received any feedback ever since it's creation and because Paper is generally bigger, I think this PR might have taken *much* longer than I'd like to. I agree however that more people could benefit from it but I'm currently unable to implement the full suggestion. When I'm a bit more familiar with how patches and such work, I might PR the full version to the Paper instead (assuming nobody else would get to it before me!). Just wanted to start with Purpur, as it seems smaller and more open for small changes like this one.

patches/api/0049-add-VaultStateChangeEvent.patch
```diff
@@ -0,0 +1,76 @@
...
++import org.bukkit.block.data.type.Vault;
++import org.bukkit.event.Cancellable;
++import org.bukkit.event.HandlerList;
++import org.bukkit.event.block.BlockEvent;
++import org.jspecify.annotations.NullMarked;
++import org.jetbrains.annotations.ApiStatus;
++import org.jetbrains.annotations.NotNull;
++
++@NullMarked
++public class VaultStateChangeEvent extends BlockEvent implements Cancellable {
```
Euphillya, line 26: You are not using cancelable, so I think it can be removed.

granny:
Hey! Sorry for the late response. This PR needs to be updated to account for the hard fork changes, which means you'd have to re-learn how to apply your change, to a degree. It's definitely a lot easier now, though.

I'd recommend pushing this PR to Paper. The maintainers are currently doing their best to stay on top of PRs, so this would be the perfect time to try and get this merged in by them.

Grabsky:
I don't know when I will be able to work on that, so if anyone wants to take this PR to Paper, feel free to do so.
This would be very appreciated. :smiley: 

Because of hard fork changes, only one per-file patch will be needed and event class can be added directly to the source.
Class to patch: `net.minecraft.world.level.block.entity.vault.VaultBlockEntity`

<br>

> **@Grabsky**: I don't know if I put the implementation in the right place because currently - I'm afraid - the event is also called when `Vault.State` is changed via API and doing so inside the event can create an indefinite loop if handled incorrectly.

Ever since I wrote this, I looked at the implementation of the mentioned method and IIRC it is setting the property manually and not calling method modified by this PR. Correct me if I'm wrong.

PR #1650, opened 2025-03-30
Fix floating warning log when the kick event is canceled

pedroagrs:
No description; the title is self-explanatory.


spattyan:
Great job!

granny:
PR was opened in upstream at https://github.com/PaperMC/Paper/pull/12374

PR #1652, opened 2025-04-12
Add check for max growth age special case

Dreeam-qwq:
If set `settings.blocks.xxx.max-growth-age` to 0, the server will crash when loaded the world (ticking fluid), since there is a `random.nextInt(bound)` in `GrowingPlantHeadBlock#getStateForPlacement`, the bound should be a int > 0, and the random range is 0 (inclusive) to bound (exclusive).

So just add a check for this special case, directly use 0, instead of calling nextInt

Based on the minecraft wiki, the vanilla growth age range for Twisting Vines, Kelp, etc, are 0~25, so I think no need to check cases of <0, just assume the user will set a value between 0 and 25 in purpur config 

---
Crash stacktrace
image


granny (approved):
LGTM! Thanks :)

PR #1656, opened 2025-04-22
Add configuration setting to re-add End void rings

linaTetris:
For players who want the End void rings to be re-added to Purpur given that Paper has patched it. Players who are interested in building large-scale map art (especially transparent map art) or building large space-based structures in the void will benefit from changing the configuration settings.

This PR adds a simple global setting switch `generate-end-void-rings` to toggle the rings (default set to `false` to have the fixed terrain generation).

**Requesting backport of this config setting into 1.21.4.**


granny:
Haven't looked at the PR yet, but if it gets accepted/merged, a backport will not happen.

granny (approved):
LGTM! Just needs a PR to PurpurDocs to document the option :)

Onako2:
> LGTM! Just needs a PR to PurpurDocs to document the option :)

If no one wants to do it, I can do it

PR #1666, opened 2025-05-27
Add a setting to retain projectile owner after hit when owner is disconnected

Lemonzyy:
After building a [wireless-activated teleporter](https://youtu.be/MxEMNUu06bk?si=tBaRy69Av7CZUxsX) on a Purpur server, my friends and I noticed that the teleporter stopped working after re-logging into the server. Interestingly, this mechanism functions correctly in Vanilla Minecraft.

Upon digging into the issue, I discovered that the `Owner` property of the arrow disappears when the owner logs out. This happens because the arrow continuously collides with an entity (specifically, a wind charge), and due to a fix implemented in Paper, the `Owner` is set to `null` when the owner is disconnected from the server.

I introduced a configuration option under `gameplay-mechanics.retain-projectile-owner-on-hit-if-owner-offline`. When enabled, this option prevents the projectile from losing its owner upon hitting an entity while the owner is offline, thereby restoring the expected behavior and ensuring compatibility with redstone contraptions that rely on projectile ownership.

### DIFF
--- a/purpur-server/minecraft-patches/features/0021-Retain-owner-after-hit-when-owner-is-disconnected.patch
+++ b/purpur-server/minecraft-patches/features/0021-Retain-owner-after-hit-when-owner-is-disconnected.patch
@@ -1,0 +1,20 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Lemonzy <lemonzy.pro@gmail.com>
+Date: Wed, 28 May 2025 01:01:10 +0200
+Subject: [PATCH] Retain owner after hit when owner is disconnected
+
+Previously, due to a fix in Paper, projectiles would lose their owner when they hit an entity while the owner was disconnected from the server. This behavior caused issues with certain redstone contraptions relying on the projectile's owner.
+
+diff --git a/net/minecraft/world/entity/projectile/Projectile.java b/net/minecraft/world/entity/projectile/Projectile.java
+index 0a1cee73ee7d895dba55745647daa4870038515c..04e28b213561cdd3410329545d0152f0c36c4195 100644
+--- a/net/minecraft/world/entity/projectile/Projectile.java
++++ b/net/minecraft/world/entity/projectile/Projectile.java
+@@ -59,7 +59,7 @@ public abstract class Projectile extends Entity implements TraceableEntity {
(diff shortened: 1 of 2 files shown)


RhythmicSys:
I don't see any major issues, granny or trouble will need to do a full review

If you haven't already, also make a PR to add the new option to the purpur documentation :3

Lemonzyy:
I'm AFK, I will do it tomorrow :)

YouHaveTrouble:
@Lemonzyy Can you link the paper patch that supposedly causes the owner to be removed?

Lemonzyy:
@YouHaveTrouble 
On paper-archive it is `patches/server/0803-Refresh-ProjectileSource-for-projectiles.patch`.

Toffikk:
Maybe you can also pr this to paper as an unsupported-setting? i think they would accept this if it breaks vanilla in some way

granny:
Looks like this was the original PR that added the lines you're changing: https://github.com/PaperMC/Paper/pull/9715

That area of the code changed in 1.21.6. Looks like the issue you were running into now happens on vanilla 1.21.6 as well (needs more testing).

granny:
Yeah, the issue you're trying to solve now exists on vanilla 1.21.6 (and 1.21.7) as well. I've found two tickets on mojira about it but no confirmation from mojang: [MC-299067](https://bugs.mojang.com/browse/MC/issues/MC-299067) & [MC-299115](https://bugs.mojang.com/browse/MC/issues/MC-299115)

Lemonzyy:
If projectile ownership loss is the correct behavior on Vanilla, I think that there's no need to add my feature. We will have to wait and see.

granny:
I feel like it's an oversight on mojang/paper tbh

I think this could still be "fixed" in a different way that doesn't change the same area of the code

granny:
This is a bit of a late reply - Has anyone been able to test if this still happens in either vanilla or paper/purpur? [MC-299067](https://bugs.mojang.com/browse/MC/issues/MC-299067) & [MC-299115](https://bugs.mojang.com/browse/MC/issues/MC-299115) have both been marked as resolved bugs by Mojang.

granny:
Closing due to inactivity. This should no longer be an issue based on my previous comment.

PR #1680, opened 2025-06-29
Fix phantom going at high speed when ridden

mxuexe:
As explained in Discord, the old code keeps leftover movement each tick, which adds up and makes the phantom go faster and faster.
- Resetting it to ```Vec3.ZERO``` which clears it and the phantom will move as much as it's told.


granny (approved):
Thanks for the PR!

PR #1682, opened 2025-06-29
Fix phantom high speed by modifying the speed

mxuexe:
As noticed, thanks to Billy !

In older versions, <1.21.6, Minecraft multiplies the vector by 0.02 and now in 1.21.6 they multiply it by 0.2

This might be a Minecraft bug, but the simplest fix (Thank to Billy again) would to multiply the speed by 0.1 to match the older versions.

This pr is working as intended (tested)


granny (approved):
I ended up reverting the 0.1 change since bringing back the friction code made it work as before.

Thanks for the original PR as well as this one!

PR #1684, opened 2025-06-30
Fix llama not moving when ridden, Fixes #1507

### LINKED ISSUE #1507: Llama are not controllable
### Spark link

https://spark.lucko.me/ECUeK9Rs15

### Expected behavior

As written in the purpur.yml config, the llama should be rideable and controllable, but when I tame it and put a carpet on it, and step on it, it is ridable but not controllable, it does not react to WASD commands

### Observed/Actual behavior

At present, the tamed and equipped llama of a carpet is only ridable but not controllable

### Steps/models to reproduce

Open purpur.yml and set llama controllable and ridable
Then, ride it

### Purpur version

git-Purpur-2169 (MC: 1.20.4)*

### Agreements

- [X] I am running the latest version of Purpur available from https://purpurmc.org/downloads.

### Other

_No response_

mxuexe:
The issue was ```AbstractHorse``` overrides ```getControllingPassenger()``` which for the llama will always return false.

The fix was to override that method in the ```Llama``` to match the other mobs, which now makes the llama ridable and moving.


granny (approved):
Thank you for figuring this out! :tada:

you've managed to fix a purpur bug that's existed since 1.20.4 :)

Onako2:
Finally we have a hero!!!

PR #1688, opened 2025-07-05
Fix mobs not burning in daylight

mxuexe:
Mobs don't burn in daylight e.g. zombies, skeletons etc... using ```/summon```

I saw creating a mob using ```/summon``` will always have the value ```shouldBurnInDay = false``` since it doesn't override it from LivingEntity

A simple fix would be to uncomment the variables in Zombie and AbstractSkeleton and set them to true, as done with phantoms (who were already working before because of that).

### DIFF
--- a/purpur-server/minecraft-patches/features/0018-API-for-any-mob-to-burn-daylight.patch
+++ b/purpur-server/minecraft-patches/features/0018-API-for-any-mob-to-burn-daylight.patch
@@ -35,7 +35,7 @@ index 34e0fbef06b0c7aededf27fe9dc64f3f6f33e3ae..ce3e5ec505ac37c820436bcf7c7d6452
          this.type = entityType;
          this.level = level;
 diff --git a/net/minecraft/world/entity/LivingEntity.java b/net/minecraft/world/entity/LivingEntity.java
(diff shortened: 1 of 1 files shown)


Tencryn:
I'm not sure why Phantom still has that property but I can confirm that with Zombie and AbstractSkeleton, `shouldBurnInDay` is correctly overwritten.

The problem arises when `readAdditionalSaveData` is called when `Purpur.ShouldBurnInDay` is not present. Why in earth this method gets called upon `/summon`, I'm not sure. It doesn't get called when a mob is spawned naturally or spawned via spawn egg.

As `shouldBurnInDay` defaults to false within `LivingEntity`, as part of the "API for any mob to burn daylight" patch, we could modify this line in `LivingEntity`:
`this.shouldBurnInDay = input.getBooleanOr("Purpur.ShouldBurnInDay", false);`
to:
`this.shouldBurnInDay = input.getBooleanOr("Purpur.ShouldBurnInDay", this.shouldBurnInDay);`

This way, we don't need to uncomment the respective properties in Zombie and AbstractSkeleton. It doesn't make sense for Phantom to still have shouldBurnInDay if it's supposed to be moved to LivingEntity anyway.

PR #1689, opened 2025-07-05
A better version to fix mobs not burning in daylight - Better version

mxuexe:
Mobs don't burn in daylight e.g. zombies, skeletons etc... using ```/summon```

This is a better version as said by Billy, I'll keep the old pr in case you want to use that one.

Thanks to **Tencryn** for the suggestion and idea !


granny (approved):
LGTM! For future reference, please make sure to add `// Purpur` comments for every change you make in a patch :)

Since the "API for any mob to burn daylight" feature is a feature patch, any change related to that feature should reside in that feature patch. I've moved the `CraftPhantom` hunk into said feature patch!

PR #1708, opened 2025-10-18
Fix fireball griefing not overriding gamerule - Fix #1702

### LINKED ISSUE #1702: 'fireballs-mob-griefing-override' not working properly. Ghasts can break blocks when set to false.
### Spark link

N/A

### Expected behavior

'fireballs-mob-griefing-override' should disable all types of Ghast griefing in every world. While other mob-griefing overrides, such as for Creepers, work properly, it doesn’t seem to work for Ghasts.

### Observed/Actual behavior

First Test: Ghasts were spawned in both the Overworld and the Nether. When fireballs-mob-griefing-override was set to false in purpur.yml, fireballs thrown by Ghasts still broke dirt and netherrack blocks. 

Second Test: However, when testing with the Vanilla Tweaks 'Anti Ghast Grief' datapack, the issue was resolved.

Third Test: After removing the Vanilla Tweaks 'Anti Ghast Grief' datapack, the same issue as in the first test occurred.

### Steps/models to reproduce

Image

### Purpur version

1.21.8-2497

### Agreements

### Other

Removing `fireballs-mob-griefing-override` and turning it into
`ghast:
  allow-griefing: true
  mob-griefing-override: default
`
would be much more user-friendly. Just an honest thought :D

madebytwentyfive:
This fixes issue #1702

- The issue was even if in the config ```fireballs-mob-griefing-override``` is set to true or false, the gamerule will still override it, since the check was never used (Variable ```_boolean```) . and ```Level.ExplosionInteraction.MOB``` will still check the gamerule, therefore override the config.

- I was able to make the check directly with the use of ```_boolean``` to see if it will grief the blocks.

Tested with every config ```default/true/false``` and works amazing!

### DIFF
--- a/purpur-server/minecraft-patches/features/0015-Add-mobGriefing-override-to-everything-affected.patch
+++ b/purpur-server/minecraft-patches/features/0015-Add-mobGriefing-override-to-everything-affected.patch
@@ -240,10 +240,10 @@ index 19211c0d78113da7827a071fd510b4f36e63a9f2..4e1a7b6caf8c0a3721a3a799261d7b63
  
      protected boolean canReplaceCurrentItem(ItemStack candidate) {
 diff --git a/net/minecraft/world/entity/projectile/LargeFireball.java b/net/minecraft/world/entity/projectile/LargeFireball.java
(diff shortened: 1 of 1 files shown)


granny:
Superceded by 7525225be708beb299d4599cf5f229b25cc7e9f5, appreciate the contribution nonetheless!

PR #1713, opened 2025-10-26
Add option to disable Warden Sonic Boom

madebytwentyfive:
Add option to disable Warden Sonic Boom.

I just added the check in the warden's conditions to use the sonic boom.


granny (approved):
LGTM! Thanks :)

PR #1714, opened 2025-10-27
Add mobs breeding chance to make twins

madebytwentyfive:
Hey!

So as I was going through discussions, I found #523 to add to mobs the option to have twins when breeding (Liked by Billy ^_^)

I managed to add this option ```breeding-twin-chance``` as a percentage to have two babies.

This affects all breedable mobs, including the villager.

Also, this is a big feature patch, so *to review carefully*  (I don't think there are any mistakes as it's pretty well tested)


DereC4:
this is such a fun vanilla+ addition i love it

PR #1718, opened 2025-11-01
Fix lagging threshold causing to slow the server

madebytwentyfive:
Calling `getTPS()` will cause to calculate the tps 4 times (Since it uses `computeTPS()` which calls `getTPS(params..)` 4 times to return a double[]).

- Since we only need the first value, then we'll avoid the `getTPS()` and instead calculate only that value (Instead of 4).

- i also added it to the tps bar

Spark: https://spark.lucko.me/v552TtyFyC

Spark with me inside the world and tpsbar enabled: https://spark.lucko.me/vUA85z47lI

I think it can a permanent or temporary fix based on paper changes (idk lol)


madebytwentyfive:
This was fixed on upstream.

PR #1726, opened 2025-11-20
Fix fireball mobgriefing override not working

### LINKED ISSUE #1722: 'fireballs-mob-griefing-override' is not working.
### Spark link

https://spark.lucko.me/G0ihkFMfBU

### Expected behavior

When setting 'fireballs-mob-griefing-override' to false or true in purpur.yml, it should override the server's mobGriefing gamerule for the fireballs only. In my case, setting 'fireballs-mob-griefing-override' to false should prevent ghasts from destroying blocks when with the fireballs.

### Observed/Actual behavior

Setting the value to false does absolutely nothing and ghasts can still grief with fireballs. Setting the value to true and setting the mobGriefing gamerule to false does allow fireballs to grief which is the expected behavior. The only way to fully disable fireball griefing is setting 'fireballs-mob-griefing-override' to false or default and just disabling mobGriefing gamerule which defeats half the purpose of the override when it can only force it true and not false.

### Steps/models to reproduce

Setup a completely fresh minecraft server running latest version of Purpur and set 'fireballs-mob-griefing-override' to false in purpur.yml.

### Purpur version

This server is running Purpur version 1.21.10-2527-HEAD@edbd95c (1970-01-01T00:00:00Z) (Implementing API version 1.21.10-R0.1-SNAPSHOT)

### Agreements

### Other

_No response_

madebytwentyfive:
The existing does so if the game rule `this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)` is true, then the whole condition is true, which is why blocks still get destroyed even when the config is set to **false**.

The fix is to check if the **source** is a `fireball`, it will check the purpur config in `this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING, this.purpurConfig.fireballsMobGriefingOverride)`.

Else it will use the default one `this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)`.

This is tested with every value and it works.

Fixes #1722

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/server/level/ServerLevel.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/server/level/ServerLevel.java.patch
@@ -201,7 +201,7 @@
              case NONE -> Explosion.BlockInteraction.KEEP;
              case BLOCK -> this.getDestroyType(GameRules.RULE_BLOCK_EXPLOSION_DROP_DECAY);
 -            case MOB -> this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)
-+            case MOB -> (source instanceof net.minecraft.world.entity.projectile.LargeFireball && this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING, this.purpurConfig.fireballsMobGriefingOverride)) || this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING) // Purpur - Add mobGriefing override to everything affected
++            case MOB -> ((source instanceof net.minecraft.world.entity.projectile.LargeFireball) ? this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING, this.purpurConfig.fireballsMobGriefingOverride) : this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)) // Purpur - Add mobGriefing override to everything affected
                  ? this.getDestroyType(GameRules.RULE_MOB_EXPLOSION_DROP_DECAY)
                  : Explosion.BlockInteraction.KEEP;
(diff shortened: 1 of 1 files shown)


BillyGalbreath:
~~This doesn't change anything. They're the same thing.~~ nvm. see [below](https://github.com/PurpurMC/Purpur/pull/1726#issuecomment-3560703554)

madebytwentyfive:
Mind if you explain ?

madebytwentyfive:
```diff
- case MOB -> (source instanceof net.minecraft.world.entity.projectile.LargeFireball && this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING, this.purpurConfig.fireballsMobGriefingOverride)) || this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING) // Purpur - Add mobGriefing override to everything affected
+ case MOB -> ((source instanceof net.minecraft.world.entity.projectile.LargeFireball) ? this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING, this.purpurConfig.fireballsMobGriefingOverride) : this.getGameRules().getBoolean(GameRules.RULE_MOBGRIEFING)) // Purpur - Add mobGriefing override to everything affected
```

- Let's take the **gamerule** set to `true` and in **config** it's set to `false`. **source** is also a `LargeFireball`:

1. Old version:
- (true && false) || true -> true, means it can grief (Which is not correct, because the override is false).

2. New version:
- true ? false : true -> false, means it can't grief blocks which respects the config.


Note that **getBoolean(key, value)** doesn't override the gamerule.
Looking at the code, it just checks if the **value** is not null: If it's not null, it will return the **value**, else it will return the **key's value**.

BillyGalbreath:
> Mind if you explain ?

my bad. as i was writing it out to explain myself, i caught the subtle difference lol

don't mind me ^_^

granny (approved):

granny:
Thank you for the PR!

PR #1728, opened 2025-11-24
Add "/afk" command.

Mickey42302:
I would like to suggest the addition of an "/afk" command. This is designed to be used with the AFK system that is built into Purpur.

Permission nodes: "bukkit.command.afk", "bukkit.command.afk.other"


granny:
This has been PR'd before: https://github.com/PurpurMC/Purpur/pull/1539

The reason it was denied applies here as well

PR #1746, opened 2026-01-04
Make beacons not apply effects when covered or base destroyed while using "allow-effects-with-tinted-glass"

### LINKED ISSUE #1741: "allow-effects-with-tinted-glass: true" effective even with solid blocks above and when beacon base blocks removed
### Spark link

https://spark.lucko.me/WxS1JDIHwP

### Expected behavior

I have 
allow-effects-with-tinted-glass: true
in purpur.yml. If you set up your beacon as normal with sky access and select the effect you want, then add tinted glass, the beacon beam is gone but the effect is still renewed as expected. Adding solid blocks above the tinted glass or removing the beacon base blocks should stop renewal of the beacon effect.

### Observed/Actual behavior

If you add solid blocks above the tinted glass, the beacon still works, and even more surprising, if you remove the beacon base blocks, it also still renews the effects.

### Steps/models to reproduce

Build a beacon as normal and select and activate an effect. 
Add tinted glass on top.
Add solid blocks above that and/or remove the base. 
The effect will still be renewed.

### Purpur version

[20:45:50 INFO]: This server is running Purpur version 1.21.10-2535-HEAD@0a2dc04 (1970-01-01T00:00:00Z) (Implementing API version 1.21.10-R0.1-SNAPSHOT)
Previous: 1.21.10-2531-b7b7ba2 (MC: 1.21.10)

### Agreements

### Other

_No response_

iamgabrieltv:
Resolves #1741 which reported unexpected beacon behavior while using tinted glass and "allow-effects-with-tinted-glass".

The problem with the base not updating was that with tinted glass being an opaque block it would still cause `blockEntity.beamSections` to be empty causing the below code to not check if the base of the beacon has changed.
```java
if (!blockEntity.beamSections.isEmpty()) { 
    blockEntity.levels = updateBase(level, x, y, z);
}
```

The problem with the opaque block check failing above tinted glass was that `isTintedGlass` served as an override to the condition before applying the effects. To fix it I just set `isTintedGlass` to false if the block is opaque and not tinted glass.

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/world/level/block/entity/BeaconBlockEntity.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/world/level/block/entity/BeaconBlockEntity.java.patch
@@ -25,14 +25,23 @@
          if (blockEntity.lastCheckY < y) {
              blockPos = pos;
              blockEntity.checkingBeamSections = Lists.newArrayList();
-@@ -195,6 +_,7 @@
+@@ -195,11 +_,15 @@
                      }
                  }
              } else {
+-                if (section == null || blockState.getLightBlock() >= 15 && !blockState.is(Blocks.BEDROCK)) {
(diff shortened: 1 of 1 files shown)


granny (approved):
LGTM! Thanks for the PR :)

PR #1750, opened 2026-01-08
Fix hardcoded obsidian check when igniting a portal

JLyne:
BaseFireBlock.isPortal currently has a hardcoded check for a neighbouring Obsidian block, which prevents portals including Crying Obsidian from being ignited if the targeted block has no neighbouring Obsidian. 

This PR replaces the check with testing against PortalShape.FRAME which respects the Crying Obsidian setting.

Before:
https://github.com/user-attachments/assets/fa2083be-35aa-484c-afca-814483aa1de2

After:
https://github.com/user-attachments/assets/94069a85-d332-4c1d-831c-8b67b1f18b92


granny (approved):
LGTM!

PR #1758, opened 2026-02-12
Fix api not working for all mobs

ghost:
- Fix the API working only on zombie, skeleton and phantom by removing the `EntityTypeTags.BURN_IN_DAYLIGHT` check, since this one only contains zombie, skeleton variants and phantom.

- After doing this, I also fixed zombie horse and zombie nautilus (Now they will burn in day light) and Parched (After fixing the API, Parched will start burning, so this makes him not burn in daylight). 

This works all correct after testing.


granny (approved):
Thanks for the PR!

PR #1759, opened 2026-02-20
Refactor BeehiveTask for readability and minor code quality improvements

Toxez:
Improved the overall readability and structure of BeehiveTask without changing
its behavior. This includes:

- Reduced duplication of plugin channel identifiers
- Added final modifiers where appropriate
- Improved buffer handling and variable naming
- Added JavaDoc to clarify the purpose of the task
- Simplified early-return checks for better flow
- Minor cleanup to make the code more consistent with Purpur style

### DIFF
--- a/purpur-api/src/main/java/org/purpurmc/purpur/event/PreBlockExplodeEvent.java
+++ b/purpur-api/src/main/java/org/purpurmc/purpur/event/PreBlockExplodeEvent.java
@@ -37,7 +37,7 @@ public class PreBlockExplodeEvent extends BlockExplodeEvent implements Cancellab
 
     @Override
     public boolean isCancelled() {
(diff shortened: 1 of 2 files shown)


Krakenied:
Is it AI?

purpur-api/src/main/java/org/purpurmc/purpur/event/PreBlockExplodeEvent.java
```diff
@@ -37,7 +37,7 @@ public float getYield() {
 
     @Override
     public boolean isCancelled() {
-        return this.cancelled;
+        return this.cancelled; // minor update
```
Krakenied, line 40: Lmao, what's that

PR #1761, opened 2026-02-28
Fix Wither not being ridable

ghost:
Remove the second `this.moveControl` since it overrides the value of the ridables config.


granny (approved):
LGTM! Thanks :)

PR #1762, opened 2026-03-04
Fix one punch man not killing mobs in some cases

ghost:
This happens if `max_health` attribute and entity's health are set higher than `9999F`.

The fix would be to always set the damage amount to the entity's health, so no matter its health, it will get killed.


granny (approved):
Thanks!

PR #1765, opened 2026-03-30
Fix multiple bugs for 26.1 release readiness

### LINKED ISSUE #1763: Axolotls begin suffocating after 15 seconds instead of vanilla's 5 minutes.
### Spark link

https://spark.lucko.me/K3lwAzEDwx

### Expected behavior

Axolotl survives outside of water for 5 minutes before taking damage

### Observed/Actual behavior

Axolotl begins taking damage after 15 seconds out of water

### Steps/models to reproduce

Give yourself axolotl spawn egg by any means

Apply moderate force to the button/key referenced in binding to "Use Item/Place Block" while standing on a block and looking down

Watch the axolotl die after a short period

### Purpur version

This server is running Purpur version 1.21.11-2566-HEAD@0fbf8d2 (2026-03-03T03:49:17Z) (Implementing API version 1.21.11-R0.1-SNAPSHOT)
Previous: 1.21.11-127-bd74bf6 (MC: 1.21.11)

### Agreements

### Other

Deviation of behavior happens between Paper and Purpur, and issue was replicated with no extra plugins. Possible raw number of seconds to tick conversion happening since the numbers line up.

dreamy32:
## Summary

- **Fix drowning air ticks overriding entity-specific values** — the `drowningAirTicks` config was blindly applied to all entities, causing axolotls to suffocate in 15 seconds instead of 5 minutes. Now only overrides entities with default air supply. Fixes #1763
- **Fix stored bee data modifications being silently discarded** — `update()` was calling `copyTagWithEntityId()` three times, writing to throwaway copies. Uses `getUnsafe()` to modify the actual underlying tag.
- **Fix DolphinSpit using wrong damage method** — `onHitEntity` was calling `hurt()` instead of `hurtServer()` with a `ServerLevel` check, matching the pattern used by `PhantomFlames`.
- **Fix WaterMoveControllerWASD using wrong jump input source** — was reading `rider.jumping` field directly instead of `getLastClientInput().jump()` like every other controller.
- **Use `onGround()` method instead of deprecated field in controllers** — updated `FlyingMoveControllerWASD`, `FlyingWithSpacebarMoveControllerWASD`, `MoveControllerWASD`, and `LookControllerWASD`.
- **Log IOException instead of silently swallowing it in PurpurConfig** — filesystem errors when loading `purpur.yml` are now logged instead of ignored.

## Test plan

- [ ] Verify axolotls survive out of water for ~5 minutes (6000 ticks) with default config
- [ ] Verify dolphins/nautiluses retain their custom air supply values
- [ ] Verify `drowningAirTicks` config still works for normal entities
- [ ] Place bees in a beehive, modify stored bee custom name/persistent data, verify changes persist
- [ ] Test dolphin spit damage on entities while riding a dolphin
- [ ] Test ridable water mobs (spacebar/jump input)
- [ ] Test ridable flying mobs (ground vs air speed)
- [ ] Corrupt or remove purpur.yml and verify error is logged on startup

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/world/entity/Entity.java.patch
@@ -38,7 +38,7 @@
      // Paper start - EAR 2
      public final boolean defaultActivationState;
      public long activatedTick = Integer.MIN_VALUE;
-@@ -443,10 +_,22 @@
+@@ -443,10 +_,26 @@
      }
      // Paper end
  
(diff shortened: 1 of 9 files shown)


ghost:
How much AI is this?

26.1 is not even released, why fixing it when it's not finished yet?

PR #1768, opened 2026-04-04
Fix axolotl dying after 300 ticks

### LINKED ISSUE #1763: Axolotls begin suffocating after 15 seconds instead of vanilla's 5 minutes.

ghost:
Fixes #1763 

This restores default value which is 6000 ticks and not using the config anymore. 

Axolotls have `6000 ticks` of air supply, so we can use the default value of `300 ticks` provided by the config. So either leave it like this, or make a custom config value for axolotls ?

Also, I made `Entity` use `getDefaultMaxAirSupply` (Used by vanilla to set maxAirTicks) since nautilus and axolotls use other names for air supply `AXOLOTL_TOTAL_AIR_SUPPLY` and `NAUTILUS_TOTAL_AIR_SUPPLY`


granny:
superceded by #1769

PR #1771, opened 2026-04-12
Update LlamaJoinCaravanEvent arguments

PureGero:
`caravanHead` was renamed to `tail` in 26.1

Without the fix in this PR, this exception is thrown:
```
[21:13:07 ERROR]: Entity threw exception at world:197.0565253594869,71.0,352.56243817392556
java.lang.NullPointerException: Cannot invoke "net.minecraft.world.entity.animal.equine.Llama.getBukkitEntity()" because "this.caravanHead" is null
	at net.minecraft.world.entity.animal.equine.Llama.joinCaravan(Llama.java:492)
```


granny (approved):
Thank you!

PR #1774, opened 2026-04-21
Persist hidden items in permission-based ender chest rows

### LINKED ISSUE #1773: Ender chest items are lost when permission-based rows are reduced
### Spark link

n/a

### Expected behavior

When `ender_chest.six_rows` and permission-based ender chest rows are enabled, removing a player’s higher row permission should only hide those rows from access. Items stored in the now-hidden slots should remain saved and become available again if the player regains the required permission.

### Observed/Actual behavior

If a player has `purpur.enderchest.rows.six`, stores items in rows 4-6, then later loses that permission, those items can be lost on the next login/save.

The ender chest container reports its size based on the player’s current row permission, and the load/save logic also uses that reported size. When the player’s permission is reduced, slots above the new allowed size are skipped during load/save and are omitted from saved EnderItems.

### Steps/models to reproduce

- /lp user USER permission set purpur.enderchest.rows.six true
- Fill up Ender Chest
- /lp user USER permission set purpur.enderchest.rows.five true
- /lp user USER permission set purpur.enderchest.rows.six false
- Logout and back in (refresh perms)
- /lp user USER permission set purpur.enderchest.rows.six true
- Logout and back in (refresh perms)
- Missing items from last row

### Purpur version

Purpur Build 2568

### Agreements

### Other

Originally reported in: https://github.com/CraftCanvasMC/Canvas/issues/200
Fixed in: https://github.com/CraftCanvasMC/Canvas/pull/201

GoodrichDev:
~~This fixes item loss with permission-based ender chest rows.~~

~~Previously, `PlayerEnderChestContainer#getContainerSize()` returned the player’s currently permitted row count, and the ender chest load/save methods also used that value. If a player stored items in higher rows and later lost the corresponding permission, those hidden slots could be skipped during load/save and omitted from saved player data.~~

~~This changes ender chest persistence to use the full backing container size while keeping the permission-limited size for access/UI behavior. Hidden rows remain inaccessible without permission, but their contents stay saved and become available again if the player regains the required permission.~~

EDIT: This adds support for preserving ender chest contents in rows hidden by permission-based row limits.

Currently, when permission-based ender chest rows are enabled, reducing a player’s allowed row count also reduces the size used by ender chest load/save logic. As a result, slots outside the player’s current permission range are not retained. This change keeps the existing permission-based access behavior intact: players can only view and interact with rows they currently have permission for. The new behavior is that hidden rows remain part of the backing storage during persistence, so their contents can be retained and become available again if the player regains the required row permission.

- Closes https://github.com/PurpurMC/Purpur/issues/1773

@/BillyGalbreath says this is not a bug and is working as intended, asked if it instead it could be intended to work this way. Hence the strike-through and rename

### DIFF
--- a/purpur-server/minecraft-patches/features/0003-Barrels-and-enderchests-6-rows.patch
+++ b/purpur-server/minecraft-patches/features/0003-Barrels-and-enderchests-6-rows.patch
@@ -84,7 +84,7 @@ index e77bfcd31cdcfd5836dc5db561e3fe2bc552a4b1..afa2ead0548766669201526d83f351c2
          return new ChestMenu(MenuType.GENERIC_9x6, containerId, inventory, container, 6);
      }
 diff --git a/net/minecraft/world/inventory/PlayerEnderChestContainer.java b/net/minecraft/world/inventory/PlayerEnderChestContainer.java
-index 4df3a32faf85595372f4b250482d852c985ea3ab..8347150af55119d772b797e79be412f7e17a0f1f 100644
+index 4df3a32faf85595372f4b250482d852c985ea3ab..b792b83a67293baaab02f5e46bad88afbfdfe0dc 100644
 --- a/net/minecraft/world/inventory/PlayerEnderChestContainer.java
 +++ b/net/minecraft/world/inventory/PlayerEnderChestContainer.java
 @@ -26,11 +26,18 @@ public class PlayerEnderChestContainer extends SimpleContainer {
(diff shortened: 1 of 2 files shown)


granny (changes requested):
Based on all the discussion around this change, I feel like it wouldn't hurt to put this behind a config option that's enabled by default. I'll make sure to include it in an announcement so that those who relied on the container size "deleting" data when the container resizes can still have that available to them. Persisting the data would be the default, since I agree that most people would expect it to persist.

GoodrichDev:
I made the necessary changes @granny

purpur-server/minecraft-patches/features/0003-Barrels-and-enderchests-6-rows.patch
```diff
@@ -107,6 +107,31 @@ index 4df3a32faf85595372f4b250482d852c985ea3ab..8347150af55119d772b797e79be412f7
...
+-            if (item.isValidInContainer(this.getContainerSize())) {
++            if (item.isValidInContainer(storageSlotCount)) {
+                 this.setItem(item.slot(), item.stack());
+             }
+         }
+     }
+ 
+     public void storeAsSlots(final ValueOutput.TypedOutputList<ItemStackWithSlot> output) {
+-        for (int i = 0; i < this.getContainerSize(); i++) {
++        int storageSlotCount = org.purpurmc.purpur.PurpurConfig.enderChestPreserveHiddenItems ? super.getContainerSize() : this.getContainerSize(); // Purpur - preserve hidden ender chest rows
```
granny:
```suggestion
+        int storageSlotCount = org.purpurmc.purpur.PurpurConfig.enderChestSixRows && org.purpurmc.purpur.PurpurConfig.enderChestPreserveHiddenItems ? 54 : this.getContainerSize(); // Purpur - Barrels and enderchests 6 rows
```

granny (approved):
LGTM! Thank you for the PR!

PR #1787, opened 2026-06-18
Fix sneak-to-bulk-process

### LINKED ISSUE #1778: `sneak-to-bulk-process` option does not work
### Spark link

N/A

### Expected behavior

a portion of the stack should be consumed and the composter should be full after one attempt

### Observed/Actual behavior

9 times out of 10 it doesn't consume a portion of the stack, and doesn't fill the composter to be full

### Steps/models to reproduce

1. set `blocks.composter.sneak-to-bulk-process` to true
2. try sneak-right-clicking a stack of vegetation into the composter

### Purpur version

```
> version
[23:37:20 INFO]: Checking version, please wait...
[23:37:20 INFO]: This server is running Purpur version 26.1.2-2577-HEAD@fe26844 (2026-04-24T00:27:53Z) (Implementing API version 26.1.2.build.2577-experimental)
```

### Agreements

### Other

_No response_

kacimiamine:
Looking at the code, `addItem()` uses a `willRaiseLevel` boolean to determine whether the composter's level should increase.

When `willRaiseLevel` is false, `addItem()` returns the original `BlockState`. meaning, `newState == state`, causing the `if` condition to be `false`.

This means sneak bulk processing only starts when the first attempt successfully raises the composter level. If the first attempt fails its compost chance, only a single item is processed instead of continuing until the composter is full.

The fix is to remove the state check, I tested it and it works as intended.

Fixes #1778

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/world/level/block/ComposterBlock.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/world/level/block/ComposterBlock.java.patch
@@ -20,7 +20,7 @@
 +            if (newState == null) {
 +                return InteractionResult.PASS;
 +            }
-+            if (level.purpurConfig.composterBulkProcess && player.isShiftKeyDown() && newState != state) {
++            if (level.purpurConfig.composterBulkProcess && player.isShiftKeyDown()) {
 +                BlockState oldState;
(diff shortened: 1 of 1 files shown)


granny (approved):
Thank you so much for the PR! Looks good to me :)

PR #1792, opened 2026-06-29
Add wind charge to projectile offset setting

kacimiamine:
Saw someone looking for this setting in discord, I figure out it should be added to other projectile settings.

Tested and works as intended.


granny (approved):
LGTM!

PR #1793, opened 2026-07-05
Use stable bossbar IDs for Purpur TPS/RAM bars

LilPhyziX:
This changes Purpur’s internal TPS, RAM, and compass bossbars to use vanilla ServerBossEvent instances with deterministic per-player IDs.

Previously, BossBarTask created a new Adventure BossBar each time a player joined a backend server. Behind proxies that skip the configuration phase, stale client-side bossbars may not be cleared during server transfer. Since each backend created a new bossbar ID, the client could stack multiple TPS/RAM bars.

Using stable IDs based on the player UUID and task type makes the bossbars idempotent across backend transfers, so the next backend updates the same client-side bossbar instead of adding another one.

P.S, Yes, I used AI. I am not a professional java developer. It was looked over by a friend of mine that knows and understands java.

### DIFF
--- a/purpur-server/src/main/java/org/purpurmc/purpur/task/BossBarTask.java
+++ b/purpur-server/src/main/java/org/purpurmc/purpur/task/BossBarTask.java
@@ -1,11 +1,15 @@
 package org.purpurmc.purpur.task;
 
 import net.kyori.adventure.bossbar.BossBar;
+import net.minecraft.server.level.ServerBossEvent;
 import net.minecraft.server.level.ServerPlayer;
+import net.minecraft.world.BossEvent;
 import org.bukkit.Bukkit;
+import org.bukkit.craftbukkit.entity.CraftPlayer;
 import org.bukkit.entity.Player;
 import org.bukkit.scheduler.BukkitRunnable;
 
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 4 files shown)


granny:
Thanks for the contribution!

This changes Purpur's internal TPS, RAM, and compass bossbars to use vanilla `ServerBossEvent` instances with deterministic per-player IDs.

While I understand the issue you're trying to solve, this is unfortunately not something we'll be merging.

The root problem described here is specific to proxy setups that intentionally bypass the normal client configuration/login flow. Purpur assumes vanilla-compatible behavior, and we generally don't add complexity to work around undefined behavior introduced by third-party proxy implementations.

This patch also significantly increases the complexity of the bossbar implementation by replacing the existing Adventure API with NMS `ServerBossEvent`s, adding conversion helpers, introducing deterministic UUID generation, and requiring CraftBukkit/NMS conversions throughout the TPS, RAM, and Compass tasks. All of those changes exist solely to support a scenario outside of Purpur's intended scope.
While the implementation itself looks reasonable, the maintenance cost and additional NMS coupling outweigh the benefit for a niche compatibility issue. This is the sort of behavior that is better addressed by the proxy software responsible for the transfer, rather than by every backend server.

Thanks again for taking the time to put this together and for being upfront about how you developed the patch. We appreciate the effort, but we'll be closing this one.

P.S, Yes, I used AI for this response.

PR #1794, opened 2026-07-05
Fix registerMinecraftDisabledCommands using wrong default value

toprakdevx:
Fixes a copy-paste bug where registerMinecraftDisabledCommands read its default from registerMinecraftDebugCommands.


kacimiamine:
You're not using the correct branch. You should pr to the branch `ver/26.2`

PR #1795, opened 2026-07-05
Add config option to disable book editing

toprakdevx:
Adds a new setting `enable-books` (default: true) in purpur.yml that allows server admins to completely disable book editing and signing. When set to false, players trying to edit books are disconnected with a clear message.

This provides a defense against book-based duplication exploits that have been used on Minecraft servers. Servers that anticipate being a target for duping can disable books entirely without needing additional plugins.


kacimiamine:
- Pointing to the wrong branch, you should use `ver/26.2`
- This pr includes the commit of the other pr you opened #1794 which is wrong
- I'm sure this can be easily done with a plugin

PR #1796, opened 2026-07-05
Fix registerMinecraftDisabledCommands using wrong default value

toprakdevx:
Fixes a copy-paste bug where registerMinecraftDisabledCommands was reading its default value from registerMinecraftDebugCommands instead of its own field. This caused the two settings to interfere with each other when toggled.

registerMinecraftDebugCommands defaults to false, so setting registerMinecraftDisabledCommands to true would still show as false on a config reload since it was reading the wrong field.


kacimiamine (approved):

kacimiamine:
Make sure to close the two other pull requests that was pointing to `ver/1.21.11`

granny (approved):

PR #1797, opened 2026-07-05
Add config option to disable book editing

toprakdevx:
Adds a new setting `enable-books` (default: true) in purpur.yml that allows server admins to completely disable book editing and signing. When set to false, players trying to edit books are disconnected with a clear message.

This provides a defense against book-based duplication exploits that have been used on Minecraft servers.


YouHaveTrouble:
What's the justification for this instead of a simple plugin that cancels PlayerEditBookEvent?

Mickey42302:
This would be much easier to implement via a plugin. Perhaps it could be a module in PurpurExtras?

PR #1798, opened 2026-07-08
[ci-skip] chore: update param names in patch to match vanilla

Toffikk:


granny (approved):
Thank you for the PR!

PR #1802, opened 2026-07-11
Add region-safe packet broadcaster

Smorki:
Add RegionPacketBroadcaster for entity tracker packet broadcasts.
Add PacketBroadcastContext for packet interception, replacement, and cancellation.
Route controllable mob look-controller packets through the entity-owned Folia scheduler.
Validation

Ran ./gradlew applyAllPatches
Ran ./gradlew :purpur-server:compileJava

### DIFF
--- a/purpur-server/src/main/java/org/purpurmc/purpur/controller/LookControllerWASD.java
+++ b/purpur-server/src/main/java/org/purpurmc/purpur/controller/LookControllerWASD.java
@@ -2,11 +2,11 @@ package org.purpurmc.purpur.controller;
 
 
 import net.minecraft.network.protocol.game.ClientboundMoveEntityPacket;
-import net.minecraft.server.level.ServerLevel;
 import net.minecraft.util.Mth;
 import net.minecraft.world.entity.Mob;
 import net.minecraft.world.entity.ai.control.LookControl;
 import net.minecraft.world.entity.player.Player;
+import org.purpurmc.purpur.network.RegionPacketBroadcaster;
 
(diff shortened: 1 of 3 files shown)


Smorki:
Apologies for the previous commit author confusion, I've rebased and fixed the configuration. Thanks for your patience

kacimiamine:
As said in the PR before

1. Why did you even set granny in the first place?

2. What is the real purpose of this as your description in the old PR looks AI generated?

3. Your validation doesn't prove anything here.

Is this AI generated ?

Smorki:
I appreciate your concern. To be completely honest, I am learning as I go and I did use AI assistance for some parts of the implementation and the PR description, as I'm still familiarizing myself with the codebase.

However, I am putting a lot of effort into understanding the logic behind the packet broadcasting and the Folia scheduler integration. I'm not trying to claim work that isn't mine, I'm genuinely trying to contribute to Purpur. If the implementation needs adjustments or if I need to prove my understanding of the packet flow, I am more than happy to do so. Thanks for the feedback.

YouHaveTrouble:
> Purpose of this PR: This is a performance-oriented optimization aimed at decoupling packet broadcasting from the main server tick loop. The use of taskScheduler was intended to ensure thread safety in environments where entity processing might be offloaded. If this approach doesn't align with the current Purpur maintenance guidelines, I am happy to revert to a standard synchronous implementation. Please let me know your preferred direction

This doesn't make any sense in the context of purpur server. Nor does it improve performance, it actually adds overhead. There will be no scenario in purpur where this is gonna make sense.

PR #1803, opened 2026-07-13
Add leaves instant decay option

kacimiamine:
Add a new option to instant decay the leaves after breaking a tree.

Note that even if `randomTick()` gets called every tick in `tick()` , it doesn't have any performance issues even if it runs on ALL leaves, since for leaves not decaying, it's just a boolean check and return.

I tested this and I didn't see any issues. This would be a nice addition since doing it with a plugin requires a bit more logic and might not be as intended.

How this works:
https://discord.com/channels/685683385313919172/720374890472931328/1525151676447723523

Update: thanks to Billy's suggestion, we made a new method instead of randomTick, so depending on the config, only one of them will fire (either decay on randomTick, or decay on tick if config is enabled)


granny (approved):
LGTM!

PR #1807, opened 2026-07-17
Optimize dolphinspit tick

Smorki:
Small optimizations in DolphinSpit#tick to reduce redundant method calls, casts, and allocations inside the tick loop.

Changes
Hoist getX/Y/Z out of the particle loop.
Hoist motDouble.x/y/z() and (ServerLevel) level() cast out of the particle loop
Merge the two setDeltaMovement calls into a single call using a local Vec3, removing the intermediate getDeltaMovement()
fetch.


Smorki:
Sorry for the noise on the previous PRs. First one was AI-generated with wrong author, and I made a mess trying to reopen with a clean branch. I've now re-forked and pushed a clean, single-commit PR here: #1808


Thanks for the patience.

PR #1808, opened 2026-07-17
Optimize DolphinSpit tick logic

Smorki:
Small optimizations in DolphinSpit#tick to reduce redundant method calls, casts, and allocations inside the tick loop.

Changes
Hoist getX/Y/Z out of the particle loop.
Hoist motDouble.x/y/z() and (ServerLevel) level() cast out of the particle loop
Merge the two setDeltaMovement calls into a single call using a local Vec3, removing the intermediate getDeltaMovement()
fetch.

### DIFF
--- a/purpur-server/src/main/java/org/purpurmc/purpur/entity/projectile/DolphinSpit.java
+++ b/purpur-server/src/main/java/org/purpurmc/purpur/entity/projectile/DolphinSpit.java
@@ -49,29 +49,38 @@ public class DolphinSpit extends LlamaSpit {
 
         this.preHitTargetOrDeflectSelf(hitResult);
 
-        double x = this.getX() + mot.x;
-        double y = this.getY() + mot.y;
-        double z = this.getZ() + mot.z;
+        final double posX = this.getX();
+        final double posY = this.getY();
+        final double posZ = this.getZ();
+
+        double x = posX + mot.x;
+        double y = posY + mot.y;
+        double z = posZ + mot.z;
 
         this.updateRotation();
 
         Vec3 motDouble = mot.scale(2.0);
+        final double motX = motDouble.x();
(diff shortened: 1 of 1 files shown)


Smorki:
Sorry for the noise from the two earlier closed PRs. I had a branch/author metadata mess while cleaning up my fork.

This one is the clean version with only the intended change.

kacimiamine:
The code is not wrong, but this doesn't provide any meaningful performance improvement. This feels more like trying to improve something that already works good enough, and it's not really worth it

Smorki:
Thanks for the feedback. My aim with this change is to minimize redundant method calls, casts, and object allocations within the tick loop. While the individual impact might seem small, hoisting these operations out of the loop reduces cumulative overhead in high-entity scenarios, which is a good practice for maintaining tick performance. @kacimiamine

kacimiamine:
My point is you're trying to improve something that doesn't need improvement at all. There is no performance issue or anything with that, so this just adds a bit of complexity for little to no practical gain.

Smorki:
Fair point regarding the complexity. I view this as a minor cleanup that aligns better with standard coding practices, in addition to the micro-optimization. I’ll leave it to the maintainers to decide if the trade-off is worth it for the codebase.

YouHaveTrouble:
What tick performance?

How many of those you think are up at any given time?
The entire projectile is removed after 20 ticks whatever happens too.

Please don't have ai look for "optimizations" this way, you're basically wasting our time for having to look at this.

If you have any benchmark of a **realistic** scenario where this is a problem that requires addressing, I'd like to see it. Otherwise please close this.

Smorki:
Fair enough I understand your point regarding the lifecycle of the projectile and the lack of a measurable impact ıll close this PR Thanks for the clarification

PR #1812, opened 2026-07-19
Fix one punch man not killing some mobs

### LINKED ISSUE #1811: [Bug] Zombie variants survive one-punch-in-creative empty hand despite receiving massive damage and having no armor or reinforcements
### Spark link

https://spark.lucko.me/8eitbnPCeY

### Expected behavior

The massive damage from `one-punch-in-creative` should kill the zombie in one hit, exactly as it does for every other mob.

### Observed/Actual behavior

The zombie correctly receives the enormous damage value (health drops to near zero) but death is not triggered. It remains alive with a sliver of health and can only be finished off by a **weapon** — non-weapon items are ineffective. This indicates the massive damage is applied, but the death condition is not satisfied unless the damage source is of a specific type associated with weapons.

### Steps/models to reproduce

1. Set `one-punch-in-creative: true` in `purpur.yml` and restart the server.  
2. Enter Creative mode with an empty hand.  
3. Summon an unarmored zombie:  
   `/summon minecraft:zombie ~ ~ ~ {ArmorItems:[{},{},{},{}]}`  
4. Punch it once — the zombie takes massive damage but survives with a tiny amount of health.  
5. Summon a zombie with disabled reinforcements:  
   `/summon minecraft:zombie ~ ~ ~ {ArmorItems:[{},{},{},{}],CanBreakDoors:0b,ReinforcementChance:0.0f}`  
6. Punch it once — it also survives despite receiving the full massive damage.  
7. (Control) Punch a pig, cow, or creeper with an empty hand — they all die in one hit.  
8. Switch to a **weapon** (e.g., sword, axe) and hit the surviving zombie — it dies instantly.  
9. Switch to a **non-weapon item** (e.g., block, food) and hit the surviving zombie — it does **not** die.

### Purpur version

This server is running Purpur version 26.2-2610-HEAD@42bdbde (2026-07-19T01:37:54Z) (Implementing API version 26.2.build.2610-stable)
Previous: 26.2-25-d0623b0 (MC: 26.2)

### Agreements

### Other

When `one-punch-in-creative` is set to `true` in `purpur.yml`, punching a zombie or its variants (husk, drowned, etc.) with an **empty hand** in Creative mode correctly deals the massive damage but does **not** kill the mob — it remains alive at a sliver of health. Other tested mobs (pigs, cows, creepers, skeletons) are all killed instantly by the same empty-hand punch.  
The tested zombies were **completely unarmored**, and the issue persisted even after setting their **reinforcement chance to 0** via NBT (`ReinforcementChance:0.0f`). This eliminates armor mitigation and the reinforcement mechanic as possible causes.

kacimiamine:
It looks like there are checks after for the absorption, armor, resistance etc... that decreases the damage amount, so using `getHealth()` won't kill the mobs with those modifiers (wither, zombie etc...)

The best fix it to set the damage to the max value possible `Float.MAX_VALUE. This is more appropriate and also safe since we have a check after the damage:
```java
if (Float.isNaN(damage) || Float.isInfinite(damage)) {
    damage = Float.MAX_VALUE;
}
```

Fixes #1811

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/world/entity/LivingEntity.java.patch
@@ -97,7 +97,7 @@
 +
 +                if (attackDamage.doubleValue() == 0.0D) {
 +                    // One punch!
-+                    damage = this.getHealth();
++                    damage = Float.MAX_VALUE;
 +                }
 +            }
 +        }


momo-OwO-qwq (approved):
good job bro 👍

granny (approved):

PR #1823, opened 2026-08-10
Add commands for managing operators

Mickey42302:
I would like to suggest implementing commands for managing operators.

Currently, there is no way to change a player's operator level without editing the "ops.json" file. Similarly, you cannot change the value of "bypassesPlayerLimit" without editing the file.

This addition would allow server owners to have more control over players they grant operator privileges to. The "/oplevel" command can be used to change the permission level, while "/bypassplayerlimit" changes the value of the "bypassesPlayerLimit" entry.

There is also an "/opinfo" command, which shows a list of operators and allows you to see information for a specific player.

Having these commands alongside a permissions plugin would allow for even better customisation of a server's permission setup. This would be especially helpful in cases where a functionality does not have a working permission node.

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/commands/Commands.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/commands/Commands.java.patch
@@ -14,7 +14,7 @@
              RaidCommand.register(this.dispatcher, context);
              DebugPathCommand.register(this.dispatcher);
              DebugMobSpawningCommand.register(this.dispatcher);
-@@ -300,6 +_,14 @@
+@@ -300,6 +_,17 @@
              StopCommand.register(this.dispatcher);
              TransferCommand.register(this.dispatcher);
              WhitelistCommand.register(this.dispatcher);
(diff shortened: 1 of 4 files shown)


kacimiamine:
This feels more like something that should be done as a plugin. This is also not useful, since all servers prefer to use permissions than giving operator access to their players.

Mickey42302:
I tried to achieve that using a plugin. However, I couldn't find a way to change the permission level or bypassesPlayerLimit with the API.

The goal of this command suite is to allow for more control over the operator privileges themselves. It is not intended to replace the use of a permissions plugin. Both of them are essential, as a number of game's features do not have a permission node. And for some of the features that do, they don't fully work as intended.

kacimiamine:
This add very little to not benefit at all. I don't see why we should have an operator managing system when production servers don't really use operators, and testing servers where op is used, don't really care who is operator and who is not. 

It looks like this is something only a few users, if none at all, will find it useful. I prefer to expand the API to allow this to be made as a plugin rather than have it as a server feature.

BillyGalbreath:
> I tried to achieve that using a plugin. However, I couldn't find a way to change the permission level or bypassesPlayerLimit with the API.

What if we (or Paper) added the missing API, instead?

Mickey42302:
I'd be happy to give making the plugin another try if the API were to be added. If the Paper Team is okay with it, it would be best to add the missing code to Paper. That way, it could benefit more projects.

As much as I would like to help out with that, my main specialty is using an API. I'm not familiar with making/expanding an API.

kacimiamine:
You can make a feature request on Paper. If it's not accepted, we will consider adding it here.

Mickey42302:
I've opened a feature request on the Paper repository.

granny:
> Having these commands alongside a permissions plugin would allow for even better customisation of a server's permission setup

A better suggestion would be for the server owner to use LuckPerms and completely disable Op in the LuckPerms config. Op should be completely avoided and is not meant to be used alongside a permissions plugin.

> This would be especially helpful in cases where a functionality does not have a working permission node.

If a certain functionality does not have a working permission node, then it'd be better to fix it than include a bandaid fix in the server software.

As @kacimiamine stated, this would be better as a plugin. I don't see these commands being implemented in Purpur or Paper since LuckPerms exists.

Linking the following feature request OP created on the Paper Repo for tracking: https://github.com/PaperMC/Paper/issues/14177

Issue #1826, opened 2026-08-25
Burp sound does not play


RoboMWM:
### Spark link

https://spark.lucko.me/RCAoUFBQ27

### Expected behavior

Burp sound should play after consuming any food item. (golden apple, steak, rotten flesh, etc.)

### Observed/Actual behavior

Burp sound does not play under any condition or configuration.

### Steps/models to reproduce

2. Consume any food item (I've tested rotten flesh, golden apple, steak)
3. No burp sound played or present in subtitles

Attempted with default configuration:

```yml
        burp-delay: 10
        burp-when-full: false
```

and

```yml
        burp-delay: 10
        burp-when-full: true
```

and

```yml
        burp-delay: 0
        burp-when-full: true
```

### Purpur version

This server is running Purpur version 26.2-2622-HEAD@48bcd91 (2026-08-20T00:17:43Z) (Implementing API version 26.2.build.2622-stable)
  * You are running the latest version
  Previous: 26.2-119-bb09b43 (MC: 26.2)

### Agreements

- [x] I am running the latest version of Purpur available from https://purpurmc.org/downloads.
- [x] I have searched for and ensured there isn't already an open issue regarding this.
- [ ] I ticked all the boxes without actually reading them
- [x] My version of Minecraft is supported by Purpur.

### Other

Burp sound plays after consumption of any food item on Paper: `This server is running Paper version 26.2-119-main@bb09b43 (2026-08-25T13:58:23Z) (Implementing API version 26.2.build.119-stable)`

PR #1828, opened 2026-08-27
Default burp delay option to 0 instead of 10 (vanilla behavior)

RoboMWM:
Vanilla behavior of playing the burp sound immediately after finishing eating (i.e. `burp-delay: 0`) was previously impossible; this PR restores the logic to play the vanilla burp sound if burp delay is <= 0. Also sets the `burp-delay` to `0` by default as per vanilla behavior.

Awaiting feedback of which method is desirable (early return vs. if/else).

Partially addresses #1826 (`burp-delay: 0` case)

Set as draft until I finish testing.

### DIFF
--- a/purpur-server/minecraft-patches/sources/net/minecraft/world/food/FoodProperties.java.patch
+++ b/purpur-server/minecraft-patches/sources/net/minecraft/world/food/FoodProperties.java.patch
@@ -1,17 +1,15 @@
 --- a/net/minecraft/world/food/FoodProperties.java
 +++ b/net/minecraft/world/food/FoodProperties.java
-@@ -42,9 +_,11 @@
+@@ -42,6 +_,12 @@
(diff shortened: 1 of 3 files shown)


RoboMWM:
Finished testing: Vanilla behavior of playing burp sound immediately after eating works when `burp-delay: 0`

All other test cases in #1826 are still valid. (As per [billy's message](https://discord.com/channels/685683385313919172/685683819608670209/1541881982085238817) I plan to remove `playerBurpWhenFull` and always set burpDelay if `burp-delay` is greater than 0 in another PR, or I can do it in this one.)

granny:
> Early return to keep the hunk concise. Basically, do whatever you can to make as little "diff changes" as possible to accomplish your objective.

Early return in this case to keep the hunk concise, but depending on what you're modifying, an if/else could be less diff. Basically, do whatever you can to make as little "diff changes" as possible to accomplish your objective.

granny (approved):

PR #1831, opened 2026-09-02
Change the default permissions for "/plugins" and "/version".

Mickey42302:
I would like to suggest changing the default value of the permission nodes for some of the built-in commands.

At present, all players can execute "/plugins". Regular players do not need access to the list of plugins a server has installed. It is better to let the server owner decide if someone should be able to see what plugins they have installed instead of granting access to it by default.

Regular players do not need access to "/version" either. It would be better to restrict this command to operators as well, just like with the "/purpur version" command.

### DIFF
--- a/purpur-api/paper-patches/files/src/main/java/org/bukkit/util/permissions/CommandPermissions.java.patch
+++ b/purpur-api/paper-patches/files/src/main/java/org/bukkit/util/permissions/CommandPermissions.java.patch
@@ -5,10 +5,10 @@
  
          DefaultPermissions.registerPermission(PREFIX + "help", "Allows the user to view the vanilla help menu", PermissionDefault.TRUE, commands);
 -        DefaultPermissions.registerPermission(PREFIX + "plugins", "Allows the user to view the list of plugins running on this server", PermissionDefault.TRUE, commands);
-+        DefaultPermissions.registerPermission(PREFIX + "plugins", "Allows the user to view the list of plugins running on this server", PermissionDefault.OP, commands);
++        DefaultPermissions.registerPermission(PREFIX + "plugins", "Allows the user to view the list of plugins running on this server", PermissionDefault.OP, commands); // Purpur - Require operator privileges to see installed plugins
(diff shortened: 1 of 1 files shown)


kacimiamine:
What is the point of hiding those? I don't see any reason at all.

And if server owners want to explicitly hide them, they can just use luckperms or something like commandwhitelist.

Mickey42302:
Because they don't need the information. When configuring servers, it is best to follow the Principle of Least Privilege.

Not having access to "/plugins" and "/version" won't stop people from playing on the server or enjoying the content it offers. It only prevents them from seeing a full list of installed plugins and what version of Purpur is installed.

In some cases, hiding the information can even help keep the server safe. For example, if an exploit has been found, denying access to "/version" will prevent a malicious user from seeing what build number is installed.

kacimiamine:
- Having access to the version and plugins also doesn't prevent them from stop playing. And as I said, if an owner doesn't want, he still can use the permission or something to hide the commands (why would you do that anyway)

- About security, security through obscurity is not security. If there's an exploit, no one cares about your list of plugins or version, they just spam those exploits and see if they work.

Mickey42302:
I'm not vouching for security through obscurity. Server owners should always keep their server up to date.

Although denying access to "/version" isn't a silver bullet, it will still prevent players from easily retrieving the version data. Thus, if a malicious user doesn't know that the server owner has a patched build number installed, they will just be wasting their time by spamming the exploit to see if it works.

So far, every server I've played on denies access to "/plugins" and "/version" too. I've yet to see any public servers which allow regular players to use the commands.

kacimiamine:
> So far, every server I've played on denies access to "/plugins" and "/version" too

A lot of servers are also proud of showing their plugins. And I think you're missing whatever I said above (which I said two times), you can still disable those with permissions.

Mickey42302:
I'm aware that you can override the permission nodes with permission plugins (such as LuckPerms). I just think that the commands shouldn't have to be restricted manually; they should be restricted to operators out of the box.

If a server owner really wants all of their players to see what plugins they are using, they will grant the "bukkit.command.plugins" permission node to their default group.

granny:
Closed for the same reasons as outlined in https://github.com/PaperMC/Paper/pull/14168

